PR #143, opened 2023-05-08
Improve logging

giovanni-guidini:
Trying to strike the right balance between verbose and non-verbose mode in the CLI.

Particularly for the upload command it's been not satisfactory. The list of files being uploaded as coverage is now listed in non-verbose mode.

Plus there's a positive feedback now that the process (command) finished - for ALL commands. So you know that at least the CLI did its thing, even though there might still be errors.

And again for the upload command, separate better the response from codecov-api and the response from the PUT request afterwards.
To help debugging and all that.

### DIFF
--- a/tests/services/commit/test_commit_service.py
+++ b/tests/services/commit/test_commit_service.py
@@ -31,10 +31,7 @@ def test_commit_command_with_warnings(mocker):
 
     out_bytes = parse_outstreams_into_log_lines(outstreams[0].getvalue())
     assert out_bytes == [
-        (
-            "info",
-            "Process Commit creating complete. --- {\"result\": \"RequestResult(error=None, warnings=[RequestResultWarning(message='somewarningmessage')], status_code=201, text='')\"}",
-        ),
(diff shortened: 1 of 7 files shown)


tests/services/commit/test_commit_service.py
```diff
@@ -73,7 +77,13 @@ def test_commit_command_with_error(mocker):
         )
 
     out_bytes = parse_outstreams_into_log_lines(outstreams[0].getvalue())
-    assert out_bytes == [("error", "Commit creating failed: Permission denied")]
+    assert out_bytes == [
+        (
+            "info",
+            "Process Commit creating complete. --- {\"result\": \"RequestResult(error=RequestError(code='HTTP Error 403', params={}, description='Permission denied'), warnings=[], status_code=403, text='Permission denied')\"}",
```
dana-yaish, line 83: can we get rid of the extra "\" in the loggings?

dana-yaish (dismissed):

dana-yaish (approved):

PR #146, opened 2023-05-11
create empty-upload command

dana-yaish:
This PR adds a command which triggers the empty-upload url, logs the result and the errors if any


codecov_cli/services/empty_upload/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+
+logger = logging.getLogger("codecovcli")
+
+
+def empty_upload_logic(commit_sha, slug, token, git_service):
+    encoded_slug = encode_slug(slug)
+    headers = get_token_header_or_fail(token)
+    url = f"https://api.codecov.io/upload/{git_service}/{encoded_slug}/commits/{commit_sha}/empty-upload"
+    sending_result = send_post_request(url=url, headers=headers)
+    log_warnings_and_errors_if_any(sending_result, "Empty Upload")
```
giovanni-guidini, line 18: I think this should come _after_ the `logger.info` message below, because that message states "Process Empty upload complete." and then adds the warnings and errors... I mean not right now, but when you update this branch it will

giovanni-guidini (commented):
Looks good in general.

Left a minor comment on user feedback order.
Also very nit, you should pick if `empty_upload` is a subservice of `upload` (so in `services/upload/empty_upload.py` and leave the test file where is) or an independent service (and move the test file).

giovanni-guidini (approved):

PR #152, opened 2023-05-16
add --url use to all commands available

dana-yaish:
Ticket: [CODE-3311](https://codecovio.atlassian.net/browse/CODE-3311?atlOrigin=eyJpIjoiODNjMGU4MTM5OTBmNDlmYmE5ZDE3YmUwMDJhYjg5ZGQiLCJwIjoiaiJ9)

[CODE-3311]: https://codecovio.atlassian.net/browse/CODE-3311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/codecov_cli/services/commit/base_picking.py
+++ b/codecov_cli/services/commit/base_picking.py
@@ -10,12 +10,13 @@ from codecov_cli.helpers.request import (
 logger = logging.getLogger("codecovcli")
 
 
-def base_picking_logic(base_sha, pr, slug, token, service):
+def base_picking_logic(base_sha, pr, slug, token, service, enterprise_url):
     data = {
         "user_provided_base_sha": base_sha,
     }
     headers = get_token_header_or_fail(token)
-    url = f"{CODECOV_API_URL}/api/v1/{service}/{slug}/pulls/{pr}"
+    upload_url = enterprise_url or CODECOV_API_URL
+    url = f"{upload_url}/api/v1/{service}/{slug}/pulls/{pr}"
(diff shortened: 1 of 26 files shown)


codecov_cli/services/commit/base_picking.py
```diff
@@ -14,7 +15,7 @@ def base_picking_logic(base_sha, pr, slug, token, service):
         "user_provided_base_sha": base_sha,
     }
     headers = get_token_header_or_fail(token)
-    url = f"https://api.codecov.io/api/v1/{service}/{slug}/pulls/{pr}"
+    url = f"{CODECOV_API_URL}/api/v1/{service}/{slug}/pulls/{pr}"
```
giovanni-guidini, line 18: Shouldn't enterprise customers also be able to pick their base?...

giovanni-guidini (approved):
Looks good. It looks like enterprise customers won't be able to base_pick, but I might have overlooked something.

I like the config options you wrote (`CODECOV_API_URL` and `LEGACY_CODECOV_API_URL`). 

I think it would be a good thing to have similar Context config, so that when writing into the context object and reading from it we use tha TAGS, not strings. I'd recommend an `Enum`. At your discretion to do it, tho.

codecov_cli/services/commit/base_picking.py
dana-yaish: you're totally right

giovanni-guidini (approved):

PR #154, opened 2023-05-17
Create plugin to compress ATS reports

giovanni-guidini:
Recently we've noticed that JSON reports with labels - the one format accepted by ATS -
are simply too big. It is likely that we are experiencing Out Of Memory errors when
trying to process them.
    
To put it in perspective, below is a comparison between reports with no labels and
with labels for the worker repo full test suite.
```
  918951 bytes (897Kb) May 15 19:45 coverage.xml # this is the regular coverage with no labels
26607438 bytes (25Mb)  May 15 19:48 label.coverage.json # this is all labels, stock
```
    
To avoid processing issues and save bandwidth and all that we will be compressing these reports.
The compression idea is simple: we build a mapping label --> int that will substitute the label itself
in the report (usually large-ish strings) for a number in the "contexts" of each file.
We then include the reverse mapping in the report (int --> label) so that we can easily parse
the compressed report when processing it.
    
This is done via the new `compress_pycoverage_contexts` plugin that can be used when uploading
with the CLI. Notice that the process is transparent for the user.
The results are promissing. Compressing the worker repo (25Mb) yields a new report with 801Kb.
    
About the new requirements.
These libs allow streaming large files (smart_open) and stream-parsing the JSON (ijson)
This means we don't ever load the full big report in memory when compressing it,
avoiding the Out Of Memory error in the plugin as well.

### DIFF
--- a/.github/workflows/push_flow.yml
+++ b/.github/workflows/push_flow.yml
@@ -39,8 +39,8 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python setup.py develop
         pip install -r requirements.txt
+        python setup.py develop
     - name: Create commit in codecov
       run: |
         codecovcli create-commit -t ${{ secrets.CODECOV_TOKEN }} --git-service github
--- a/codecov_cli/plugins/__init__.py
+++ b/codecov_cli/plugins/__init__.py
@@ -4,6 +4,7 @@ from importlib import import_module
 
 import click
 
+from codecov_cli.plugins.compress_pycoverage_contexts import CompressPycoverageContexts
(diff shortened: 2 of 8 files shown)


giovanni-guidini:
I forgot to mention it in the PR opening comment, but the results are promissing.
Check below a comparison between the mentioned reports and the new compressed one.
We can see it's even slightly smaller than the XML one (with no labels)
```
897K May 15 19:45 coverage.xml
801K May 16 11:25 label.coverage.codecov.json # compressed report
 25M May 15 19:48 label.coverage.json
```

dana-yaish (dismissed):

dana-yaish (approved):

PR #155, opened 2023-05-17
Enable other runners (DAN runner)

giovanni-guidini:
We had created a DAN runner previously, but there was no option in the label analysis
command to actually select different runners.

Another issue is that thwe return of `collect_tests` in a runner needs to be a list,
and the DAN runner was not returning that.

The changes here should make the DAN runner a viable alternative to the
PythonStandardRunner, in case we need that.

### DIFF
--- a/codecov_cli/commands/labelanalysis.py
+++ b/codecov_cli/commands/labelanalysis.py
@@ -32,6 +32,7 @@ logger = logging.getLogger("codecovcli")
     cls=CodecovOption,
     required=True,
 )
+@click.option("--runner-name", "--runner", "runner_name", help="Runner to use")
 @click.pass_context
 def label_analysis(
     ctx: click.Context,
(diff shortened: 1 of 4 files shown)


codecov_cli/commands/labelanalysis.py
```diff
@@ -32,6 +32,7 @@
     cls=CodecovOption,
     required=True,
 )
+@click.option("--runner-name", "--runner", "runner_name", help="Runner to use")
```
dana-yaish, line 35: can we provide what's the default runner we use in case the user doesn't specify one?

codecov_cli/runners/dan_runner.py
```diff
@@ -10,10 +11,19 @@
 class DoAnythingNowConfigParams(dict):
```
dana-yaish, line 11: Are the ConfigParams classes between different runners have the same behavior/properties? if so can we have a Base class for them?

giovanni-guidini: No, they are independent.

The idea of the runners is similar to the plugins, in that every plugin is (and should be) completely independent and self-contained.

I don't think we should try to impose any interface on the runner's config. Each has its own needs. The details of those needs are irrelevant from the CLI's perspective.

dana-yaish (approved):

PR #163, opened 2023-05-23
add definition lines to js analyzer

dana-yaish:
adding definition lines, which is a list of line numbers that a method, function or class is defined and the number of following lines that's part of the function, method etc

ticket: [CODE-3136](https://codecovio.atlassian.net/browse/CODE-3136?atlOrigin=eyJpIjoiNjhlZmJlOTAxMmNkNGZhZWE3OTdmMjMzMjI1ZjgzNDIiLCJwIjoiaiJ9)

[CODE-3136]: https://codecovio.atlassian.net/browse/CODE-3136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samples/outputs/sample_004.json
```diff
@@ -154,7 +154,53 @@
...
+                18,
+                0
+            ],
+            [
+                21,
+                0
+            ],
+            [
+                23,
+                0
```
dana-yaish, line 189: if you look at sample_004.js you'd find that line 23 is `function a () { function b () {} function *c () {} class D {} return }`. It has multiple functions and one class, but the definition_lines only has one [23,0] because it's a set. Should the output contain four of [23,0] to represent that it has multiple functions here or the goal for the definitions list here is different?

giovanni-guidini: That's an interesting question. We can see that this information is missing data for us to identify _what_ is being declared _where_. Even if we have the 4 `[23,0]` it's not very useful because we can't link it back to anything, really.

Added to that is the fact that this is not even used anywhere at this point: https://github.com/codecov/worker/blob/5e352be2189b197547c5370d7c70a707bc20c447/services/static_analysis/single_file_analyzer.py#L54 , and I don't really know what this would be used for, really.

However if we want a list with _all_ declaration lines, then we should include _all_ declaration lines. I would change the output to contain all the declarations (so 4 [23, 0]). That is my opinion, but there's no right or wrong in this case.

What do you think?

giovanni-guidini (commented):
Looks good in general. Let's decide what to do with multi-declarations in the same line.

giovanni-guidini (approved):

PR #173, opened 2023-06-02
Add new ignore paths

eliatcodecov:
When reviewing the ignore paths in the Node uploader, I found a few that weren't present here. I've updated the file and folder ignore paths to rectify this.

### DIFF
--- a/codecov_cli/services/upload/coverage_file_finder.py
+++ b/codecov_cli/services/upload/coverage_file_finder.py
@@ -58,7 +58,7 @@ coverage_files_excluded_patterns = [
     "*.exe",
     "*.ftl",
     "*.gif",
-    ".gitignore",
+    ".git*",
     "*.gradle",
     "*.gz",
     "*.h",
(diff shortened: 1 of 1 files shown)


codecov_cli/services/upload/coverage_file_finder.py
```diff
@@ -58,7 +58,7 @@
     "*.exe",
     "*.ftl",
     "*.gif",
-    ".gitignore",
+    ".git*",
```
eliatcodecov, line 61: ensures that .gitmodules, .gitattributes, and .gitignore are all exlcuded instead of just .git

codecov_cli/services/upload/coverage_file_finder.py
```diff
@@ -127,23 +127,29 @@
...
+    ".circleci",
     "conftest_*.c.gcov",
+    ".egg-info*",
     ".env",
     ".envs",
     ".git",
+    ".go",
     ".hg",
+    ".map",
+    ".marker",
```
eliatcodecov, line 146: .go, .map, and .marker were all seen in NodeJS, but not found here.

codecov_cli/services/upload/coverage_file_finder.py
```diff
@@ -127,23 +127,29 @@
...
     ".tox",
     ".venv",
     ".venvs",
     ".virtualenv",
     ".virtualenvs",
+    ".yarn",
```
eliatcodecov, line 152: taken from Node uploader

PR #174, opened 2023-06-06
use --file option to let users set specific files to upload

dana-yaish:


codecov_cli/services/upload/coverage_file_finder.py
```diff
@@ -163,26 +166,76 @@
...
 
-        coverage_files_paths = search_files(
-            self.project_root,
-            default_folders_to_ignore + self.folders_to_ignore,
-            filename_include_regex=regex_patterns_to_include,
-            filename_exclude_regex=regex_patterns_to_exclude,
-        )
-
-        return [UploadCollectionResultFile(path) for path in coverage_files_paths]
+        return result_files + user_result_files
```
giovanni-guidini, line 230: You should consider using sets instead of lists to avoid having duplicated paths returned (for example if a user lists a file that is found in the search it would be present in both lists and returned twice)

codecov_cli/services/upload/coverage_file_finder.py
```diff
@@ -163,26 +166,76 @@
...
+                        extra_log_attributes=dict(
+                            files=files_excluded_but_user_includes
+                        )
+                    ),
+                )
+            regex_patterns_to_include = globs_to_regex(user_filenames_to_include)
+            multipart_include_regex = globs_to_regex(
+                [str(path.resolve()) for path in self.explicitly_listed_files]
+            )
+            user_coverage_files_paths = search_files(
```
giovanni-guidini, line 204: Should we also log files that were listed by the user (explicated_listed_files) that were NOT found by this?

giovanni-guidini (commented):
Looks good in general. Left a comment about using sets.
I feel that `find_coverage_files` is getting a bit big though... maybe it would be interesting to break it up in `find_explicitly listed_files` and `_find_coverage_files` for example. Makes it easier to read and to test.

codecov_cli/services/upload/coverage_file_finder.py
dana-yaish: good point

giovanni-guidini (dismissed):
This is blocking a customer so we're deploying as is. It has been tested locally, but unit tests will follow soon.

eliatcodecov (dismissed):

PR #181, opened 2023-06-15
Fix custom plugin system config fields 

joseph-sentry:
This PR implements a fix for the incorrect custom plugin path config field, by removing the incorrect field and replacing it with a field for the module and a field for class.

This PR also implements a fix for handling cases where the custom plugin params field is missing or empty.  

Fixes: #60

### DIFF
--- a/codecov_cli/plugins/__init__.py
+++ b/codecov_cli/plugins/__init__.py
@@ -45,10 +45,13 @@ def _load_plugin_from_yaml(plugin_dict: typing.Dict):
         )
         return NoopPlugin()
     try:
-        return class_obj(**plugin_dict["params"])
+        if "params" not in plugin_dict or plugin_dict["params"] is None:
+            return class_obj()
+        else:
+            return class_obj(**plugin_dict["params"])
     except TypeError:
         click.secho(
-            f"Unable to instantiate {class_obj} with parameters {plugin_dict['params']}",
+            f"Unable to instantiate {class_obj} with provided parameters { plugin_dict.get('params', '') }",
             err=True,
         )
         return NoopPlugin()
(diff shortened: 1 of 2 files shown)


codecov_cli/plugins/__init__.py
```diff
@@ -30,18 +30,28 @@ def select_preparation_plugins(cli_config: typing.Dict, plugin_names: typing.Lis
...
+        return NoopPlugin()
+    except AttributeError:
+        click.secho(
+            f"Unable to dynamically load class {plugin_dict['class']} from module {plugin_dict['module']}",
             err=True,
         )
         return NoopPlugin()
     try:
-        return class_obj(**plugin_dict["params"])
+        if "params" not in plugin_dict or plugin_dict["params"] is None:
```
giovanni-guidini, line 48: [nit] what you wrote is perfectly fine. This is simply a suggestion for being more concise. You could do `params = plugin_dict.get("params", None)` and change the considitonal to `if params:` (and reuse `params` in the `else` too)

tests/plugins/test_instantiation.py
```diff
@@ -11,38 +11,77 @@
 
 
 def test_load_plugin_from_yaml(mocker):
-    class SamplePlugin(object):
-        def __init__(self, banana, other):
-            self.something = banana
-            self.other = other
+    class SampleModule(object):
```
giovanni-guidini, line 14: [nit] I know this is being mocked anyway, but a Module is not exactly a class.
I think that it would make more sense in the test if you do `SampleModule = mocker.MagicMock(SamplePlugin=SamplePlugin)`. 

This is because it doesn't lead people reading this code in the future to imagine that `module_obj` _should_ be a class

giovanni-guidini (approved):
Looking good overall. I'm leaving 2 nit comments, mostly on the test setup. But that's nothing that would prevent this PR from being merged, so it's up to you if you'd like to make the changes or merge this directly.

joseph-sentry:
@giovanni-guidini I made the changes!

giovanni-guidini (approved):

PR #182, opened 2023-06-19
Call static analysis finish endpoint after uploading files

joseph-sentry:
This commit changes the static-analysis command in the CLI, such that it calls the staticanalysis/analyses/<external_id>/finish endpoint in the API after it uploads the necessary files. This endpoint schedules the static analysis task on the worker immediately.

Fixes: #66

### DIFF
--- a/codecov_cli/services/staticanalysis/__init__.py
+++ b/codecov_cli/services/staticanalysis/__init__.py
@@ -140,26 +140,12 @@ async def run_analysis_entrypoint(
     else:
         logger.info("All files are already uploaded!")
     try:
-        external_id = response_json["external_id"]
-        logger.debug(
-            "Sending finish signal to let API know to schedule static analysis task",
-            extra=dict(extra_log_attributes=dict(external_id=external_id)),
-        )
-        upload_url = enterprise_url or CODECOV_API_URL
-        response = requests.post(
-            f"{upload_url}/staticanalysis/analyses/{external_id}/finish",
-            headers={"Authorization": f"Repotoken {token}"},
-        )
-        if response.status_code >= 500:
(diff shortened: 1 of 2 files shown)


codecov_cli/services/staticanalysis/__init__.py
```diff
@@ -139,6 +139,31 @@ async def run_analysis_entrypoint(
         )
     else:
         logger.info("All files are already uploaded!")
+    try:
```
dana-yaish, line 142: for readability reasons, can we extract this and put it in a function, maybe "send_finish_signal()"?

codecov_cli/services/staticanalysis/__init__.py
```diff
@@ -139,6 +139,31 @@ async def run_analysis_entrypoint(
...
+        if response.status_code >= 400:
+            raise click.ClickException(
+                f"There is some problem with the submitted information.\n{response_json.get('detail')}"
+            )
+    except requests.RequestException:
+        raise click.ClickException(click.style("Unable to reach Codecov", fg="red"))
+    logger.info(
+        "Received response from server",
+        extra=dict(
+            extra_log_attributes=dict(time_taken=response.elapsed.total_seconds())
```
dana-yaish, line 164: I think it makes sense if we logged out the response status code and and the response text, wdyt?

joseph-sentry: makes sense!

codecov_cli/services/staticanalysis/__init__.py
joseph-sentry: yes

dana-yaish (approved):
LGTM!

PR #187, opened 2023-06-26
upload process wrapper 

dana-yaish:
wrap create-commit, create-report and do-upload in one command "upload-process"
This PR adds fail-on-error option to commit and report commands. It doesn't make a difference when the command is used alone, but is useful when used in the upload-process command, which involves calling 3 commands, if one of them fails, then it doesn't call the other ones


codecov_cli/commands/upload_process.py
```diff
@@ -0,0 +1,283 @@
...
+@click.option(
+    "--git-service",
+    cls=CodecovOption,
+    fallback_field=FallbackFieldEnum.git_service,
+    type=click.Choice(service.value for service in GitService),
+)
+@click.option(
+    "--parent-sha",
+    help="SHA (with 40 chars) of what should be the parent of this commit",
+)
```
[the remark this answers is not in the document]
giovanni-guidini: There are some suggestions in this [stackOverflow](https://stackoverflow.com/questions/40182157/shared-options-and-flags-between-commands) question.

I'd say the one I was most interested in is [this one](https://stackoverflow.com/a/69082528). But creating a class is also interesting, something like a `BaseCommand`, for example.

I think we would benefit from having some level of shared options

giovanni-guidini (approved):

codecov_cli/commands/upload.py
```diff
@@ -17,169 +17,183 @@ def _turn_env_vars_into_dict(ctx, params, value):
     return dict((v, os.getenv(v, None)) for v in value)
 
 
+_global_upload_options = [
```
giovanni-guidini, line 20: This is quite nice. I think however that we could break this up even more. My suggestion would be to have a `base_cli_command_option` possibly defined in `commands/__init__.py` or some helper file. These could include the `--sha`, `--token`, `--dry-run`, `--fail-on-error` options.

Perhaps this is better suited for another round of changes if we decide to do it, so I'll leave just as a comment. What do you think? I guess we'd have to add support for some of these options to different commands...

giovanni-guidini (approved):

codecov_cli/commands/upload.py
dana-yaish: sounds like a reasonable change. I'll make a ticket for this

PR #191, opened 2023-07-06
Send labels to Codecov only after collecting them

giovanni-guidini:
These changes alter the way label analysis command works by making 2 requests to Codecov instead of 1.

Why would we want to do this, if it's more time to make a second request?
Because it takes (usually) longer to collect the labels than it takes to make a second request.
Specially for bigger repos, collecting the labels can take dozens of seconds.
By making the label collection parallel with the labels processing we can speed up the total ATS time.

We still PATCH the labels in case the worker hasn't finished the calculations, and it will
correctly refresh the request labels from the DB (based on [here](https://github.com/codecov/worker/pull/1210)).
However even if it misses the requested labels we recalculate the `absent_labels`, if needed (`_potentially_calculate_absent_labels`).

So we still guarantee the exact same response everytime. Potentially faster.

PS.: The file `test_instantiation.py` changes because of the linter

### DIFF
--- a/codecov_cli/commands/labelanalysis.py
+++ b/codecov_cli/commands/labelanalysis.py
@@ -90,17 +90,28 @@ def label_analysis(
         raise click.ClickException(
             click.style("Unable to run label analysis", fg="red")
         )
-    upload_url = enterprise_url or CODECOV_API_URL
-    url = f"{upload_url}/labels/labels-analysis"
-    token_header = f"Repotoken {token}"
 
     codecov_yaml = ctx.obj["codecov_yaml"] or {}
     cli_config = codecov_yaml.get("cli", {})
+    # Raises error if no runner is found
     runner = get_runner(cli_config, runner_name)
     logger.debug(
         f"Selected runner: {runner}",
         extra=dict(extra_log_attributes=dict(config=runner.params)),
     )
+
+    upload_url = enterprise_url or CODECOV_API_URL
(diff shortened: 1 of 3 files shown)


giovanni-guidini:
We can still make this behavior optional by adding some flag to enable/disable it.
It gives customers more option to fine-tune their CLI process.

codecov_cli/commands/labelanalysis.py
```diff
@@ -90,73 +90,68 @@ def label_analysis(
...
+
+    if eid:
+        # Initial request with no labels was successful
+        # Now we PATCH the labels in
+        patch_url = f"{upload_url}/labels/labels-analysis/{eid}"
+        _patch_labels(payload, patch_url, token_header)
+    else:
+        # Initial request with no labels failed
+        # Retry it
+        eid = _send_labelanalysis_request(payload, url, token_header)
```
dana-yaish, line 132: I can see that we patch labels in the first trial, but we don't do that if the first one failed and we retried?

giovanni-guidini: Exactly. If the first request failed, then Codecov haven't started the label analysis task yet. So instead of simply patching the labels we need to send that POST request again. Since we already have the labels at this point we can send them in the request, like it was done before.

dana-yaish: ah got it!

dana-yaish (approved):

Issue #189, opened 2023-07-06
Dependency versions should be specified in `setup.py`


scott-codecov:
For example, we depend on `click>=8.0.0` and saw an example of a customer who must have already had an older version installed:

```
  File "/usr/local/lib/python3.11/site-packages/codecov_cli/fallbacks.py", line 27, in get_default
    res = super().get_default(ctx, call=call)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Option.get_default() got an unexpected keyword argument 'call'
```

(`call` argument was introduced in 8.0.0)

`requirements.txt` specifies exact versions but that's more for development purposes.  We should move the version specs to `setup.py` and then include a `.` line in `requirements.txt` which will include all the dependencies specified in `setup.py`.  Additional dev-only dependencies like `pytest` can be left in `requirements.txt` only.

PR #192, opened 2023-07-07
Pin dependencies in setup.py and separate dev deps

joseph-sentry:
This PR pins the dependencies versions in setup.py. These dependency versions are generated by pip-compile.

This PR separates the dev dependencies into a separate dev.txt file. This file is generated by using the `pip-compile` command on the `dev.in` file.

This PR introduces a script to generate the list of dependencies in `requirements.txt` with  their versions pinned to be copied into the `setup.py` file.

This PR modifies the push GHA to use `dev.txt` instead of `requirements.txt` to install dependencies

Fixes: https://github.com/codecov/codecov-cli/issues/189

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,10 +14,18 @@ setup(
     packages=find_packages(exclude=["contrib", "docs", "tests*"]),
     description="Codecov Command Line Interface",
     long_description=long_description,
-    long_description_content_type='text/markdown',
+    long_description_content_type="text/markdown",
     author="Codecov",
     author_email="support@codecov.io",
-    install_requires=["click", "requests", "PyYAML", "tree_sitter", "httpx", "pytest", "pytest-cov", "ijson", "smart-open"],
+    install_requires=[
+        "click==8.*",
+        "httpx==0.23.*",
+        "ijson==3.*",
+        "pyyaml==6.*",
+        "responses==0.21.*",
+        "smart-open==6.*",
+        "tree-sitter==0.20.*",
+    ],
     entry_points={
         "console_scripts": [
             "codecovcli = codecov_cli.main:run",
--- a/requirements.in
+++ b/requirements.in
@@ -1,11 +1,5 @@
-click
+.
 pytest
 pytest-cov
 pytest-mock
 pytest-asyncio
-pyyaml
-responses
-httpx
-tree_sitter
-ijson
-smart-open


requirements.txt
```diff
@@ -69,10 +43,6 @@ sniffio==1.3.0
     #   anyio
     #   httpcore
     #   httpx
-tomli==2.0.1
-    # via
-    #   coverage
-    #   pytest
 tree-sitter==0.20.1
     # via -r requirements.in
 urllib3==1.26.10
```
scott-codecov, line 48: You could remove this dependency (and others) from `setup.py` since it's a transitive dependency (which you know since it says `via {other_package}`

setup.py
```diff
@@ -14,10 +14,28 @@
     packages=find_packages(exclude=["contrib", "docs", "tests*"]),
     description="Codecov Command Line Interface",
     long_description=long_description,
-    long_description_content_type='text/markdown',
+    long_description_content_type="text/markdown",
     author="Codecov",
     author_email="support@codecov.io",
-    install_requires=["click", "requests", "PyYAML", "tree_sitter", "httpx", "pytest", "pytest-cov", "ijson", "smart-open"],
+    install_requires=[
```
scott-codecov, line 20: I think these should just be our direct dependencies:

```
click
pyyaml
responses
httpx
tree_sitter
ijson
smart-open
```

joseph-sentry:
> Not sure I follow the need for `requirements.in/txt` vs `dev.in/txt`. Can what you currently have in `dev.in` be our `requirements.in`? When would we need to run `generate_reqs.py` again in the future? I was thinking we'd just update what's in `setup.py` since those should only be direct dependencies (not transitive dependencies). Ideally those versions in `setup.py` would also be looser (i.e. something like `click==8.1.x`).

My thought process was that since we are using `pip-compile` for `requirements.txt` that the deps in `setup.py` should match that. The reason for having `dev.in` and `requirements.in` is because of the use of `pip-compile`. `generate_reqs.py` is there to take the output of `pip-compile` that is the `.txt` file and make it easy to copy that into `setup.py`. I guess I thought the idea was to take whatever we currently have in `requirements.txt` and put that into `setup.py`. I think I get now that we want just loose version requirements that match up loosely with what we currently have in `requirements.txt`

scott-codecov (approved):

PR #195, opened 2023-07-12
Complete upload command

dana-yaish:
creating new command that tells codecov that the user has finished uploading and now wants to get notifications - notifications is part of the api endpoint, not related to this change specifically 
side note: I feel that 'upload-completion' is not a very descriptive name to what we're trying to do here. Do you think we can name it in a more descriptive way? Let me know if you have better ones

### DIFF
--- a/codecov_cli/commands/upload_completion.py
+++ b/codecov_cli/commands/upload_completion.py
@@ -1,0 +1,68 @@
+import logging
+import typing
+import uuid
+
+import click
+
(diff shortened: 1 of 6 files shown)


giovanni-guidini (approved):
LGTM.
Indeed `upload_completion` might not be the best... maybe trigger-notifications (because that's ultimately what we want to do). But I suppose if you think "we're letting Codecov know we completed our uploads" is not too bad.
I also think it would be better to have consistency between the api endpoint and the CLI command.

PR #201, opened 2023-07-14
Fix missing reqs pytest and pytest-cov

joseph-sentry:


matt-codecov (approved):
thanks for hopping on this fix so quickly!

PR #210, opened 2023-07-20
Retry requests on `ConnectionError` and `Timeout`

giovanni-guidini:
Adds a retry logic to `send_post_request` and `send_put_request` to retry on ConnectionError and Timeout. Also makes the legacy upload sender use those functions.

There might be a better way to do that without the custom logic, but the `Retry` option I saw you specify retry error codes, and I'm not sure ECONNRESET gives an error code. But apparently it is a `requests.exceptions.COnnectionError`, so...

Fixes uploader #537

### DIFF
--- a/codecov_cli/helpers/request.py
+++ b/codecov_cli/helpers/request.py
@@ -1,5 +1,6 @@
 import logging
 import uuid
+from time import sleep
 
 import requests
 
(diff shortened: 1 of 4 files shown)


codecov_cli/helpers/request.py
```diff
@@ -1,24 +1,46 @@
...
 
 from codecov_cli.types import RequestError, RequestResult
 
 logger = logging.getLogger("codecovcli")
 
+MAX_RETRIES = 3
 
+
+def backoff_time(curr_retry):
+    return 2 ** (curr_retry - 1)
```
scott-codecov, line 15: This will return 1, 2, and 4 seconds for the 3 retries.  Is that a helpful backoff or should it grow faster (i.e. `5 ** n` or something)?  Alternatively we could just sleep for a fixed 3 seconds or something every time.

scott-codecov (approved):

codecov_cli/helpers/request.py
giovanni-guidini: Technically it returns 0.5, 1 and 2 seconds (because the first value is 0).
I wasn't sure what to put here either, honestly... But I think it's fine, specially with timeout errors, because there's the time of the request as well.
We can also increase the number of retries to 5

giovanni-guidini: Being an iteractive tool I don't think we should use too big of a backoff period

PR #211, opened 2023-07-23
Enable ATS in the CLI

giovanni-guidini:
Changing tests that used the `-v` option (invoking the CLI) to instead use the fixture. This prevents tests from messing up the global env across tests when running pytest using ATS. As we've been discovering recently, it's a big mess sometimes.

The removal of "MASKED" from some git tests is because it was throwing errors in CI. Maybe it's a python 3.11.x thing... But I suppose they changed when the removal of secrets run. I changed for bogus secrets.

THe git services was also an error, so I had to put them in a list.

### DIFF
--- a/tests/commands/test_invoke_labelanalysis.py
+++ b/tests/commands/test_invoke_labelanalysis.py
@@ -164,7 +164,9 @@ class TestLabelAnalysisCommand(object):
         assert result.exit_code != 0
         assert "Base and head sha can't be the same" in result.output
 
-    def test_invoke_label_analysis(self, get_labelanalysis_deps, mocker):
+    def test_invoke_label_analysis(
+        self, get_labelanalysis_deps, mocker, use_verbose_option
+    ):
         mock_get_runner = get_labelanalysis_deps["mock_get_runner"]
         fake_runner = get_labelanalysis_deps["fake_runner"]
         collected_labels = get_labelanalysis_deps["collected_labels"]
@@ -203,7 +205,7 @@ class TestLabelAnalysisCommand(object):
             cli_runner = CliRunner()
             result = cli_runner.invoke(
                 cli,
-                ["-v", "label-analysis", "--token=STATIC_TOKEN", "--base-sha=BASE_SHA"],
+                ["label-analysis", "--token=STATIC_TOKEN", "--base-sha=BASE_SHA"],
                 obj={},
             )
             assert result.exit_code == 0
@@ -287,7 +289,7 @@ class TestLabelAnalysisCommand(object):
         assert str(exp.value) == "Failed to get list of labels to run"
 
     def test_fallback_collected_labels_covecov_500_error(
-        self, get_labelanalysis_deps, mocker
+        self, get_labelanalysis_deps, mocker, use_verbose_option
     ):
         mock_get_runner = get_labelanalysis_deps["mock_get_runner"]
         fake_runner = get_labelanalysis_deps["fake_runner"]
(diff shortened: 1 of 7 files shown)


giovanni-guidini:
I am fairly confident that all tests are passing now. However there's 1 more issue that I don't quite know how to solve. In the actions runs you can see many times ERROR `INTERNALERROR> sqlite3.OperationalError: no such table: context`

This is an issue with `pytest-cov` and the `cov-context` option. To figure that out I tested locally with a helper script that should be the same command ATS uses. Follow these steps to try and reproduce:

1. Run `pytest -q -collect-only > labels.txt`
2. Run the tests using the script below
```python
import subprocess

default_options = ["python", "-m", "pytest", "-v","--cov=./", "--cov-context=test"]
# default_options = ["python", "-m", "pytest", "-v", "--cov=./"]


with open("labels.txt", 'r') as fd:
    all_labels = fd.readlines()

command_array = default_options + list(
        set(
            label.split("[")[0] if "[" in label else label for label in map(str.strip , all_labels)
        )
    )

with open("debug.txt", "w") as fd:
    subprocess.run(
        command_array,
        stdout=fd,
    )
```

For me, when I run with `--cov` it works sometimes, but fails I'd say a good 60% of times. If you run without that flag it passes every time.


By running with the `-v` option I noticed that it always errors in the same test:
```
tests/commands/test_invoke_upload.py::test_upload_raise_Z_option PASSED  [100%]
tests/commands/test_invoke_upload.py::test_upload_raise_Z_option ERROR   [100%]
```

Notice that the test is repeated, and it passed the 1st time. The list of labels we passed to `pytest` only has that label once. So I guess the test is passing, then when trying to write the coverage it errors and that causes all future tests to error.

giovanni-guidini:
It's the isolated filesystem 😮‍💨

tests/commands/test_invoke_labelanalysis.py
```diff
@@ -203,7 +205,7 @@ def test_invoke_label_analysis(self, get_labelanalysis_deps, mocker):
             cli_runner = CliRunner()
             result = cli_runner.invoke(
                 cli,
-                ["-v", "label-analysis", "--token=STATIC_TOKEN", "--base-sha=BASE_SHA"],
+                ["label-analysis", "--token=STATIC_TOKEN", "--base-sha=BASE_SHA"],
```
scott-codecov, line 208: Just so I understand here...

Calling `-v` would configure the global logger to be verbose and then the rest of the tests would be run in verbose mode as well?  Whereas using the fixture just configures the logger for this test?

tests/commands/test_invoke_upload.py
```diff
@@ -32,9 +32,14 @@ def test_upload_raise_Z_option(mocker):
     upload_sender = mocker.patch.object(
         UploadSender, "send_upload_data", return_value=result
     )
+    upload_collector = mocker.patch.object(UploadCollector, "generate_upload_data")
```
scott-codecov, line 35: I think this makes sense to mock anyway.  What was previously generating the fake test coverage data?

scott-codecov (approved):

tests/commands/test_invoke_labelanalysis.py
giovanni-guidini: Pretty much, yes. I don't know exactly how that works, but I'm sure there was some overlapping across the invokes and the other unit tests... some residual configuration from them was left in the env and it would log in verbose mode when it shouldn't.

But the fixture configures just that test and then returns it to the default level (INFO). So that test can run in verbose mode in an isolated fashion.

tests/commands/test_invoke_upload.py
scott-codecov: OK, interesting.  So it was probably searching the filesystem of `codecov-cli`'s CI or something for it's own test coverage

PR #212, opened 2023-07-24
hide token in debug log

dana-yaish:
we shouldn't log out the token, fixing it

### DIFF
--- a/codecov_cli/commands/base_picking.py
+++ b/codecov_cli/commands/base_picking.py
@@ -61,7 +61,7 @@ def pr_base_picking(
             extra_log_attributes=dict(
                 pr=pr,
                 slug=slug,
-                token=token,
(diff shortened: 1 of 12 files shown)


codecov_cli/commands/base_picking.py
```diff
@@ -61,7 +61,7 @@ def pr_base_picking(
             extra_log_attributes=dict(
                 pr=pr,
                 slug=slug,
-                token=token,
+                token="NOTOKEN" if not token else (str(token)[:1] + 18 * "*"),
```
giovanni-guidini, line 64: can we move this to a util function? Or even put it in the `logging_util` that makes the formatting for the logs. We can check if any of the keys is token and hide it there.

dana-yaish: good point

giovanni-guidini (approved):

PR #215, opened 2023-07-24
Use version specified in codecov_cli/__init__.py

joseph-sentry:
__version__ in `init.py` was not updated for a while, this commit updates the version number specified by `__version__` and changes `setup.py` to use the version number specified by `__version__`


dana-yaish (approved):

setup.py
```diff
@@ -8,9 +8,12 @@
 with open(path.join(here, "README.md"), encoding="utf-8") as f:
     long_description = f.read()
 
+with open(path.join(here, "codecov_cli/__init__.py"), encoding="utf-8") as f:
+    version = f.read().split('"')[1]
```
webknjaz, line 12: @joseph-sentry FYI, you don't need such hacks. Instead, it's possible to move most of this file into a `setup.cfg` and only keep the truly dynamic bits in `setup.py`. This hugely improves maintainability of packaging.

joseph-sentry: hi @webknjaz  if I were to move some stuff over to `setup.cfg` what would be the best/idiomatic way to have one place where we set the version and have it be accessible for our code to read? [This pr](https://github.com/codecov/codecov-cli/pull/216) has a bit more context as to why I made this change

PR #217, opened 2023-07-25
Improve label analysis messages

giovanni-guidini:
In particular we will now include the `external_id` of the label analysis request object (so it's easier to find in the logs),
and the processing errors that happened in label analysis task. So the user can better self-support.

I also moved the message about the labels to run from the python runner to the command in general because it's something the user would like to see in general.


dana-yaish (approved):
LGTM

PR #219, opened 2023-07-25
Fix integrity checking readme

spectralblu:
Use cli.codecov.io instead of uploader.codecov.io as example for integrity checking the binary.

When I was integrating this CLI into our CI pipeline, I ran into this error that caused me lots of confusion, since I ended up copy/pasting from the documentation into my CI code. Hopefully this saves someone else the headache.


drazisil-codecov (approved):
Looks good! Apologies for the confusion and thanks for the PR.

PR #223, opened 2023-07-28
Add fail-on-error flag to commands that were missing it

joseph-sentry:
- Creates options.py in commands/helpers
- Moves common options to global_options
- Adds global_options decorators to some subcommands

Fixes: https://github.com/codecov/codecov-cli/issues/208

### DIFF
--- a/codecov_cli/helpers/options.py
+++ b/codecov_cli/helpers/options.py
@@ -1,0 +1,52 @@
+import click
+
+from codecov_cli.fallbacks import CodecovOption, FallbackFieldEnum
+from codecov_cli.helpers.git import GitService
+
+_global_options = [
+    click.option(
+        "-C",
+        "--sha",
+        "--commit-sha",
+        "commit_sha",
+        help="Commit SHA (with 40 chars)",
+        cls=CodecovOption,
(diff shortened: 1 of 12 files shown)


codecov_cli/helpers/options.py
```diff
@@ -0,0 +1,52 @@
...
+        "--fail-on-error",
+        "fail_on_error",
+        is_flag=True,
+        help="Exit with non-zero code in case of error",
+    ),
+    click.option(
+        "--git-service",
+        cls=CodecovOption,
+        fallback_field=FallbackFieldEnum.git_service,
+        type=click.Choice(service.value for service in GitService),
```
giovanni-guidini, line 28: please change this to `click.Choice([service.value for service in GitService])` otherwise some tests will fail ATS.

Why? I don't know exactly, but if you look on the test `test_upload_process_options` you can see that the list is empty. However if you run the command locally the list will not be empty. But wrapping it in a list first will solve the issue.

giovanni-guidini (commented):
Looking good 👍 
Left 1 comment, but in general I think this is a nice change

giovanni-guidini (approved):

Issue #207, opened 2023-07-19
CLI labels the "report_fixes" section of the upload as "path_fixes"


drazisil-codecov:

trent-codecov:
After discussion, we need to preserve backwards compat. Make the API accept both formats and change `path_fixes` to `report_fixes`. Once enough time passes we can remove the bridge.

PR #224, opened 2023-07-28
Change path_fixes to report_fixes

joseph-sentry:
Fixes: https://github.com/codecov/codecov-cli/issues/207

Depends on: https://github.com/codecov/worker/pull/58

### DIFF
--- a/codecov_cli/services/upload/upload_sender.py
+++ b/codecov_cli/services/upload/upload_sender.py
@@ -87,7 +87,7 @@ class UploadSender(object):
     ) -> bytes:
         network_files = upload_data.network
         payload = {
-            "path_fixes": {
+            "report_fixes": {
                 "format": "legacy",
                 "value": self._get_file_fixers(upload_data),
             },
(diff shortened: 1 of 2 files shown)


joseph-sentry:
Hm, it looks like the renaming will have to happen in the worker as well. @drazisil-codecov pointed out that what is happening  in `_get_file_fixers` is not [path fixing](https://docs.codecov.com/docs/fixing-paths), it's [report fixing](https://docs.codecov.com/docs/fixing-reports)

dana-yaish:
but I can see that the legacy way to get and build path fixes in worker is the same in the CLI, is this a matter of naming only?

joseph-sentry:
This shouldn't be a breaking change because the worker now handles the field being named "path_fixes" and "report_fixes" in https://github.com/codecov/worker/pull/58

joseph-sentry:
> @joseph-sentry is there any update on this?

This is ready for another review

giovanni-guidini (dismissed):

scott-codecov (approved):

PR #235, opened 2023-08-15
Update __init__.py

giovanni-guidini:
Prep for new release. In this one one of the commands was renamed (`send-notifications`) so I believe this warrants a change to 0.2.0.


dana-yaish (approved):
Do we have a specific way for numbering releases? If so, can you document it somewhere?

giovanni-guidini:
> Do we have a specific way for numbering releases? If so, can you document it somewhere?

We don't have anything to follow as of yet. I try to (very) loosely follow https://semver.org/#semantic-versioning-200, but we should probably spend some time to decide what we're gonna follow and actually follow it. Especially as more people use the CLI.

PR #237, opened 2023-08-18
use raw ref for PR branch to avoid confusion bw upstream and fork

matt-codecov:
same as https://github.com/codecov/uploader/pull/1132 but for `codecovcli`

when we upload a report for CI in a PR, we associate the report with the destination repo but we take the branch name from the source repo. so if i make a PR from my fork's `main` branch to an upstream repo's `main` branch, we upload the coverage report for the upstream repo's `main` branch and update the coverage with junk from my un-merged PR

this PR instead will upload `refs/pull/#/merge` as the branch for PRs. it's slightly less nice for same-repo PRs because in order to populate coverage for your feature branches you'll need CI running on push as well as PR, but avoids invalidating coverage for projects that use a super common fork/merge workflow

docs on the environment variables used here: https://docs.github.com/en/actions/learn-github-actions/variables

### DIFF
--- a/codecov_cli/helpers/ci_adapters/github_actions.py
+++ b/codecov_cli/helpers/ci_adapters/github_actions.py
@@ -66,22 +66,11 @@ class GithubActionsCIAdapter(CIAdapterBase):
         return os.getenv("GITHUB_REPOSITORY")
 
     def _get_branch(self):
-        branch = os.getenv("GITHUB_HEAD_REF")
-        if branch:
-            return branch
-
-        branch_ref = os.getenv("GITHUB_REF")
-
-        if not branch_ref:
-            return None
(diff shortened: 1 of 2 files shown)


dana-yaish (commented):
Thank you for solving this bug, but I have a few questions
If I understand correctly, If a PR is opened from the fork repo to the upstream one, the sent slug is the upstream's repo slug, so the bug happens because of the mistaken slug, and the coverage report gets uploaded to the upstream's branch not the fork one. How is changing the branch to be refs/pull/#/merge distinguishes between the two repos?

arosen93:
Thank you so much, @matt-codecov. This has been bugging me for ages! 😄

PR #240, opened 2023-08-28
use older ubuntu runner to have wider support

LeoQuote:
The current binary would refuse to run in debian buster and bullseye , the reason is that the glibc used is too new.

changing runner to an older one can extend the supported platform, as is documented in `pyinstaller` doc.

https://pyinstaller.org/en/v5.13.1/usage.html#making-gnu-linux-apps-forward-compatible

Upload a Build Artifact is optionally, it could be useful when debugging ci, I can remove it if you want.

This is the binary I got from build, you can check it from https://github.com/LeoQuote/codecov-cli/actions/runs/6002025929 or download it -> 
[artifact.zip](https://github.com/codecov/codecov-cli/files/12456229/artifact.zip)

I intended to switch macos build in older runner but it seems not successful, I'll leave it to someone more familliar with macos.

### DIFF
--- a/.github/workflows/release_flow.yml
+++ b/.github/workflows/release_flow.yml
@@ -42,7 +42,7 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          - os: macos-latest
+          - os: macos-11
             TARGET: macos
             CMD_BUILD: >
               STATICCODECOV_LIB_PATH=$(find build/ -maxdepth 1 -type d -name 'lib.*' -print -quit | xargs -I {} sh -c "find {} -type f -name 'staticcodecov*' -print -quit | sed 's|^./||'") &&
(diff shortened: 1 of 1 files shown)


.github/workflows/release_flow.yml
```diff
@@ -81,6 +81,10 @@ jobs:
       run: pip install pyinstaller
     - name: Build with pyinstaller for ${{matrix.TARGET}}
       run: ${{matrix.CMD_BUILD}}
+    - name: Upload a Build Artifact
```
giovanni-guidini, line 84: I think this is redundant here since all artifacts generated as made available as a release asset.
Can you please remove this?

But it was a good idea. I think we'll use this in some sort of incremental release. That will be a different context than this workflow tho

giovanni-guidini (approved):
Thanks 👍

PR #246, opened 2023-09-05
fix: JS analyser do-statement

giovanni-guidini:
We had a call with a customer trying ATS today and they had an error in the JS analyser.
The error was "first_statement referenced before assignment".
This would happen if the first statement in a `do-while` block was not a `statement_block` or `expression_statement`
(I added this case in `sample_004.js` to test it)


giovanni-guidini:
Ah would you look at that. Because of the change in `sample_004.js` (which shows the error) the static-analysis action for this PR fails with the exact same error haha

dana-yaish (approved):

PR #251, opened 2023-09-11
chore: add default config

giovanni-guidini:
Adding `codecov.yml` as the default config file the CLI will look for.
It's easy for users to forget to add that in the command, and have it misconfigured and then be like
"why is this not working?".

It's also likely that they'll put the config in the codecov.yml where the rest of they codecov config already lives.

context codecov/engineering-team#397


webknjaz:
@giovanni-guidini previously, the config of `.codecov.yml` was found automatically. Now, it silently doesn't work.

giovanni-guidini:
Hey @webknjaz, indeed that was the case.

Thankfully this PR https://github.com/codecov/codecov-cli/pull/368 addresses this issue and is included in the latest version of the CLI.
Apologies for the disruption in your workflow 😅

Issue #238, opened 2023-08-21
No support for pytest-cov 4.x.x?


Andrew-S-Rosen:
The `setup.py` is pinned to 3.* for `pytest-cov`. Is this intentional? `pytest-cov` 4.0.0 was released in Sep 2022, so I wanted to just check there isn't a way to loosen this restriction.

giovanni-guidini:
This should be supported on version [v0.3.0](https://github.com/codecov/codecov-cli/releases/tag/v0.3.0)

PR #252, opened 2023-09-11
chore: Relax pytest-cov requirement

giovanni-guidini:
context: codecov/codecov-cli#238

I suspect the restriction was there because it came from `requirements.txt` at some point in time.
I did simply used `>=` (which should include support for `4.*` and above. I did test that locally and it seem to work for some commands,
so I guess it's fine.

The big changes to `requirements.txt` were... unintentional?
I ran `pip-compile requirements.in` as usual. I suppose it doesn't matter much, as the "real"
installing is controlled by `setup.py` after all.

EDIT
I have since discovered that you should run `pip-compile --output-file=requirements.txt requirements.in setup.py` (including `setup.py` there). This makes it work better.


joseph-sentry (approved):
lgtm

PR #254, opened 2023-09-13
fix: display output if pytest fails in collection

giovanni-guidini:
If pytest fails during collection phase we were previously not showing its output
That happened because we were capturing it and not displaying
This leads to users not knowing why pytest failed during collection.

These changes surface the stdout AND stderr of pytest in the collection phase
in case of errors.

closes codecov/engineering-team#463


dana-yaish (approved):
thank you for doing this

PR #255, opened 2023-09-13
chore: remove include_curr_dir option

giovanni-guidini:
The `include_curr_dir` option is only for strict_mode in python standard runner.
Few people will ever use the strict mode.
Realistically, no one will ever make `include_curr_dir` False because it increases the change
that ATS won't work (and it doesn't work on enough siturations to add one more).

I'm keeping the `strict_mode`, but making the `include_curr_dir` always True.
The default was already True, but now there's no choice.


dana-yaish (dismissed):
LGTM

dana-sentry (approved):

dana-yaish (approved):

PR #256, opened 2023-09-13
fix: make --dry-run useful

giovanni-guidini:
This makes the --dry-run in label-analysis actually do something useful.
It prints the tests you need to run _with_ the tool options to make ats run correctly.

In the case of pytest this means that it prints `--cov-context=test` (we need this to get labels)
and then the list of labels to run, spearated by spaces.

I couldn't find a way to actually export an ENV variable safely so I'm just printing to stdout
The small problem is that there's other output with it, so I decided to add "ATS_TESTS_TO_RUN="
to make grep-ing easier. Just for good measure throw it in a file as well.

So in thory if you get ATS_TESTS_TO_RUN from your CI you can do
`pytest <YOUR NORMAL OPTIONS> <ATS_TESTS_TO_RUN>` and that should work to run all the ATS tests.

closes codecov/engineering-team#470

### DIFF
--- a/codecov_cli/commands/labelanalysis.py
+++ b/codecov_cli/commands/labelanalysis.py
@@ -157,7 +157,7 @@ def label_analysis(
             if not dry_run:
                 runner.process_labelanalysis_result(request_result)
             else:
-                _dry_run_output(LabelAnalysisRequestResult(request_result))
+                _dry_run_output(LabelAnalysisRequestResult(request_result), runner)
             return
         if resp_json["state"] == "error":
             logger.error(
@@ -285,20 +285,20 @@ def _send_labelanalysis_request(payload, url, token_header):
     return eid
 
 
-def _dry_run_output(result: LabelAnalysisRequestResult):
-    logger.info(
-        "Not executing tests because '--dry-run' is on. List of labels selected for running below."
+def _dry_run_output(
+    result: LabelAnalysisRequestResult, runner: LabelAnalysisRunnerInterface
+):
+    labels_to_run = list(
+        set(
+            result.absent_labels
+            + result.global_level_labels
+            + result.present_diff_labels
+        )
     )
-    logger.info("")
-    logger.info("Label groups:")
-    logger.info(
-        "- absent_labels: Set of new labels found in HEAD that are not present in BASE"
-    )
-    logger.info("- present_diff_labels: Set of labels affected by the git diff")
-    logger.info("- global_level_labels: Set of labels that possibly touch global code")
-    logger.info("- present_report_labels: Set of labels previously uploaded")
-    logger.info("")
-    logger.info(json.dumps(result))
+    output = runner.dry_run_runner_options + sorted(labels_to_run)
+    with open("ATS_TESTS_TO_RUN", "w") as fd:
+        fd.write(" ".join(output) + "\n")
+    click.echo(f"ATS_TESTS_TO_RUN=\"{' '.join(output)}\"")
 
 
 def _fallback_to_collected_labels(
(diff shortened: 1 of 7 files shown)


codecov_cli/commands/labelanalysis.py
```diff
@@ -285,20 +285,20 @@ def _send_labelanalysis_request(payload, url, token_header):
...
+            + result.global_level_labels
+            + result.present_diff_labels
+        )
     )
-    logger.info("- present_diff_labels: Set of labels affected by the git diff")
-    logger.info("- global_level_labels: Set of labels that possibly touch global code")
-    logger.info("- present_report_labels: Set of labels previously uploaded")
-    logger.info("")
-    logger.info(json.dumps(result))
+    output = runner.dry_run_runner_options + sorted(labels_to_run)
```
thomasrockhu-codecov, line 298: is the `--cov-context=test` necessary?

codecov_cli/commands/labelanalysis.py
```diff
@@ -285,20 +285,20 @@ def _send_labelanalysis_request(payload, url, token_header):
...
+    with open("ATS_TESTS_TO_RUN", "w") as fd:
+        fd.write(" ".join(output) + "\n")
+    click.echo(f"ATS_TESTS_TO_RUN=\"{' '.join(output)}\"")
```
thomasrockhu-codecov, line 301: it would be really nice from a consumer perspective to receive just a comma-separated list

codecov_cli/commands/labelanalysis.py
giovanni-guidini: it is *absolutely* necessary that users run pytest with this. Other tools will probably have their own quirks.

I don't want to take the chance of not including these options and just tell users to add them on their own. It's too much of a chance that someone will forget to do it and then ATS won't work.

codecov_cli/commands/labelanalysis.py
giovanni-guidini: I don't see how a comma-separated list is better than a space-separated list. Especially if you consider that command line args are a space separated list, doing it comma separated seems more work.

If you are referring to `ATS_TESTS_TO_RUN` I added that to make it easier to `grep` that line in the output. There will be other output lines, so it makes it easier to find the right one.

scott-codecov: Maybe we just output to stdout by default but can send to a file with `--ouput-path={path}` or something?  The hardcoded filename feels a little weird to me here.

codecov_cli/commands/labelanalysis.py
scott-codecov: `--cov-context=test` is what tells coveragepy to track which tests cover which line?

giovanni-guidini: Yes, that is correct.

codecov_cli/commands/labelanalysis.py
giovanni-guidini: hum... that's a valid point. If they specify `--output-path={path}` I guess we can also infer that it's `--dry-run` so users don't have to pass both options, right?

scott-codecov: I guess I was thinking you'd specify both just in case there's another command that uses `--output-path` for something else someday.

giovanni-guidini: Well options are per command so even if another command uses that some day it won't matter much. 
If it needs to be tied to `dry-run` I think it should be `dry-run-output-path`
Thoughts on naming?

thomasrockhu-codecov (approved):
Pending on @scott-codecov, but looks great!

codecov_cli/commands/labelanalysis.py
scott-codecov: I like the explicitness of 2 flags personally but I think either way is fine.

scott-codecov (approved):

scott-codecov (approved):

Issue #243, opened 2023-09-01
codecovcli static-analysis fails with RecursionError


schinckel:
I have a fairly complex django project, and when I try to use `codecovcli static-analysis`, I immediately see the following:

      $ codecovcli static-analysis
      info - 2023-09-01 15:19:46,715 -- ci service found: local
      info - 2023-09-01 15:19:48,909 -- Running the analyzer on 177419 files
      Analyzing files  [------------------------------------]    1%
      multiprocessing.pool.RemoteTraceback: 
      """
      Traceback (most recent call last):
        File "/Users/matt/.asdf/installs/python/3.9.16/lib/python3.9/multiprocessing/pool.py", line 125, in worker
          result = (True, func(*args, **kwds))
        File "/Users/matt/Development/.pypoetry/virtualenvs/datamine-33JY8UuI-py3.9/lib/python3.9/site-packages/codecov_cli/services/staticanalysis/__init__.py", line 185, in analyze_file
          output = analyzer.process()
        File "/Users/matt/Development/.pypoetry/virtualenvs/datamine-33JY8UuI-py3.9/lib/python3.9/site-packages/codecov_cli/services/staticanalysis/analyzers/javascript_es6/__init__.py", line 97, in process
          "complexity_metrics": self._get_complexity_metrics(body_node),
        File "/Users/matt/Development/.pypoetry/virtualenvs/datamine-33JY8UuI-py3.9/lib/python3.9/site-packages/codecov_cli/services/staticanalysis/analyzers/general.py", line 33, in _get_complexity_metrics
          "max_nested_conditional": self._get_max_nested_conditional(body_node),
        File "/Users/matt/Development/.pypoetry/virtualenvs/datamine-33JY8UuI-py3.9/lib/python3.9/site-packages/codecov_cli/services/staticanalysis/analyzers/general.py", line 20, in _get_max_nested_conditional

        ... a hundred repeats later ...

        File "/Users/matt/Development/.pypoetry/virtualenvs/datamine-33JY8UuI-py3.9/lib/python3.9/site-packages/codecov_cli/services/staticanalysis/analyzers/general.py", line 21, in <genexpr>
          (self._get_max_nested_conditional(x) for x in node.children), default=0
      RecursionError: maximum recursion depth exceeded

This is a closed source project, is there anything I can do to help you troubleshoot this?

thomasrockhu-codecov:
@rohan-at-sentry

rohan-at-sentry:
[...]

PR #258, opened 2023-09-14
fix: make max_conditional iterative

giovanni-guidini:
context: codecov/codecov-cli#243

User ran into `RecursionError: maximum recursion depth exceeded` running
`_get_max_nested_conditional`.

By removing the recursion from the function we hopefully fix this issue.
So now it's using text-book definition of BFS to go about finding the max depth.

Notice we do have to go through all nodes of the function because we don't know if a node that is not
of type conditional doesn't have another conditional inside of it (think block nodes, for example).
And we only increase the level when we pass through one conditional, not when we go to the next child.

close codecov/codecov-cli#243

### DIFF
--- a/codecov_cli/services/staticanalysis/analyzers/general.py
+++ b/codecov_cli/services/staticanalysis/analyzers/general.py
@@ -22,7 +22,7 @@ class BaseAnalyzer(object):
         Uses BFS to avoid recursion calls (so we don't throw RecursionError)
         """
         nodes_to_visit = deque()
-        nodes_to_visit.append([head, 0])
+        nodes_to_visit.append([head, int(head.type in self.condition_statements)])
         max_nested_depth = 0
 
         while nodes_to_visit:
--- a/codecov_cli/services/staticanalysis/analyzers/general.py
+++ b/codecov_cli/services/staticanalysis/analyzers/general.py
@@ -1,4 +1,5 @@
 import hashlib
+from collections import deque
 
 
 class BaseAnalyzer(object):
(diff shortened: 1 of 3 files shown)


codecov_cli/services/staticanalysis/analyzers/general.py
```diff
@@ -16,10 +17,26 @@ def _count_elements(self, node, types):
...
+        while nodes_to_visit:
+            curr_node, curr_depth = nodes_to_visit.popleft()
+            max_nested_depth = max(max_nested_depth, curr_depth)
+            # Here is where the depth might change
+            # If the current node is a conditional
+            is_curr_conditional = curr_node.type in self.condition_statements
+
+            # Enqueue all child nodes of the curr_node
+            for child in curr_node.children:
+                nodes_to_visit.append([child, curr_depth + is_curr_conditional])
```
matt-codecov, line 37: what should the max depth of this code be:
```python
if foo:
    for c in foo.bar:
        if c.baz:
            print('hello')
```

i think the old code + new code both handle it the same way, just confirming it's what we want. as written, the `for` statement does not halt the tree traversal, it just doesn't increase the depth when you traverse it as it's not a conditional. maybe worth adding a clarifying comment?

codecov_cli/services/staticanalysis/analyzers/general.py
```diff
@@ -16,10 +17,26 @@ def _count_elements(self, node, types):
...
-    def _get_max_nested_conditional(self, node):
-        return (1 if node.type in self.condition_statements else 0) + max(
-            (self._get_max_nested_conditional(x) for x in node.children), default=0
-        )
+    def _get_max_nested_conditional(self, head):
+        """Iterates over all nodes in a function body and returns the max nested conditional depth.
+        Uses BFS to avoid recursion calls (so we don't throw RecursionError)
+        """
+        nodes_to_visit = deque()
+        nodes_to_visit.append([head, 0])
```
matt-codecov, line 25: i think you want
```
nodes_to_visit.append([head, 1 if head.type in self.condition_statements else 0])
```

thinking of:
```
if foo:
    pass
```
old code would return 1, new code would return 0 unless you change it

matt-codecov (approved):
i think a minor change is needed, else looks great!

codecov_cli/services/staticanalysis/analyzers/general.py
giovanni-guidini: Ah yes, you're right. Thanks :)

dana-yaish (approved):

PR #267, opened 2023-09-26
fix: Fix python analysis for function docstrings

giovanni-guidini:
context: https://l.codecov.dev/hfGkNI

This is a fix for a bug that I stumbled upon today.
After much investigation it boils down to this:
    - treesitter count's docstrings as strings, not comments (which is to be expected I suppose)
    - pytest counts them as comments. They won't appear in the coverage report (if function docstring)
    - if you select a docstring as the line of surety ancestorship, because it's not in the report, it has
    no labels associated with it, so ATS selects no tests for that line

We were doing that (I believe) and selected a function docstring as the line in base to look for tests.
No tests were found, and so ATS passed, but all tests failed.

These changes fix this behavior by doing 2 things:
1. Making sure function docstrings are not in `statements` on the static analysis
2. Making sure we don't select function docstrings as the line of surety ancestorship for the actual
fist line of code in a funciton. It should be `null` instead.

**NOTES FOR THE REVIEWER**

If you'd like verify my claims for yourself, you can make use of these 2 pieces of code I created to manually test things.

The code below tests `samples/input/sample_005.py`.
You can use it to generate a coverage report with contexts to verify what is there and what isn't (`pytest --cov-. --cov-context=test test_sample_5.py && coverage json --show-contexts`)
```python
# test_sample_5.py
from samples.inputs.sample_005 import well_documented_function, less_documented_function, commented_function, MyClass

def test_well_documented_function():
    assert well_documented_function(10, 10) == (20, 0, 100)

def test_less_documented_function():
    assert less_documented_function(10, 10) == (20, 0, 100)

def test_commented_function():
    assert commented_function(10, 10) == (20, 0, 100)

def test_myclass_init():
    my_instance = MyClass()
    assert my_instance.owner == 'me'
```

You can use this little piece of code to check the output of the static analysis for the sample file
```python
import json
from pathlib import Path
from codecov_cli.services.staticanalysis import analyze_file
from codecov_cli.services.staticanalysis.types import FileAnalysisRequest


request = FileAnalysisRequest(result_filename='sample_005.py', actual_filepath=Path('samples/inputs/sample_005.py'))
result = analyze_file({}, request)
print(json.dumps(result.result))
```

### DIFF
--- a/codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
+++ b/codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
@@ -15,6 +15,9 @@ class NodeVisitor(object):
         Pytest doesn't include them in the report, so I don't think we should either,
         at least for now.
         """
+        # Docstrings have type 'expression_statement
+        if node.type != "expression_statement":
+            return False
         # Docstrings for a module are OK - they show up in pytest result
         # Docstrings for a class are OK - they show up in pytest result
         # Docstrings for functions are NOT OK - they DONT show up in pytest result
@@ -25,9 +28,7 @@ class NodeVisitor(object):
         parent_is_block = node.parent.type == "block"
         first_exp_in_block = node.prev_named_sibling is None
         is_in_function_context = (
-            node.parent.parent.type == "function_definition"
-            if parent_is_block and node.parent is not None
-            else False
+            parent_is_block and node.parent.parent.type == "function_definition"
         )
 
         return (
@@ -56,9 +57,8 @@ class NodeVisitor(object):
                 "for_statement",
                 "while_statement",
             ):
-                if node.type == "expression_statement" and self._is_function_docstring(
-                    node
-                ):
+                if self._is_function_docstring(node):
+                    # We ignore these
                     return
                 closest_named_sibling_not_comment_that_is_in_statements = (
                     self._get_previous_sibling_that_is_not_comment_not_func_docstring(
--- a/codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
+++ b/codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
@@ -10,6 +10,42 @@ class NodeVisitor(object):
         for c in node.children:
             self.visit(c)
 
+    def _is_function_docstring(self, node):
+        """Skips docstrings for funtions, such as this one.
+        Pytest doesn't include them in the report, so I don't think we should either,
+        at least for now.
+        """
+        # Docstrings for a module are OK - they show up in pytest result
+        # Docstrings for a class are OK - they show up in pytest result
+        # Docstrings for functions are NOT OK - they DONT show up in pytest result
+        # Check if it's docstring
+        has_single_child = len(node.children) == 1
+        only_child_is_string = node.children[0].type == "string"
+        # Check if is the first line of a function
+        parent_is_block = node.parent.type == "block"
+        first_exp_in_block = node.prev_named_sibling is None
+        is_in_function_context = (
+            node.parent.parent.type == "function_definition"
+            if parent_is_block and node.parent is not None
+            else False
+        )
+
+        return (
+            has_single_child
+            and only_child_is_string
+            and parent_is_block
+            and first_exp_in_block
+            and is_in_function_context
+        )
+
+    def _get_previous_sibling_that_is_not_comment_not_func_docstring(self, node):
+        curr = node.prev_named_sibling
+        while curr is not None and (
+            curr.type == "comment" or self._is_function_docstring(curr)
+        ):
+            curr = curr.prev_named_sibling
+        return curr
+
     def do_visit(self, node):
         if node.is_named:
             current_line_number = node.start_point[0] + 1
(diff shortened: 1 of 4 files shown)


codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
```diff
@@ -10,6 +10,42 @@ def visit(self, node):
...
+        has_single_child = len(node.children) == 1
+        only_child_is_string = node.children[0].type == "string"
+        # Check if is the first line of a function
+        parent_is_block = node.parent.type == "block"
+        first_exp_in_block = node.prev_named_sibling is None
+        is_in_function_context = (
+            node.parent.parent.type == "function_definition"
+            if parent_is_block and node.parent is not None
+            else False
+        )
```
matt-codecov, line 31: i had to read this a couple times to follow it. the `==` expression in the if-true clause of the ternary threw me for a loop, and i think the `node.parent is not None` check is implicitly covered in `parent_is_block` (it would error there if `parent` doesn't exist, right?)

i think you can one-line this:
```
is_in_function_context = parent_is_block and node.parent.parent.type == "function_definition"
```

codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
```diff
@@ -10,6 +10,42 @@ def visit(self, node):
...
+
+        return (
+            has_single_child
+            and only_child_is_string
+            and parent_is_block
+            and first_exp_in_block
+            and is_in_function_context
+        )
+
+    def _get_previous_sibling_that_is_not_comment_not_func_docstring(self, node):
```
matt-codecov, line 41: is a function docstring a "named" node? what does "named" mean?

codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
```diff
@@ -20,9 +56,21 @@ def do_visit(self, node):
                 "for_statement",
                 "while_statement",
             ):
-                if node.prev_named_sibling:
+                if node.type == "expression_statement" and self._is_function_docstring(
+                    node
+                ):
```
matt-codecov, line 61: could you move that expression type check inside `_is_function_docstring()`?

codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
```diff
@@ -20,9 +56,21 @@ def do_visit(self, node):
...
+                    return
+                closest_named_sibling_not_comment_that_is_in_statements = (
+                    self._get_previous_sibling_that_is_not_comment_not_func_docstring(
+                        node
+                    )
+                )
+                if closest_named_sibling_not_comment_that_is_in_statements:
```
matt-codecov, line 68: when would this be `None`? does this being `None` mean that `current_line_number` really doesn't have an ancestor?

matt-codecov (dismissed):
accepting for agility but left some style nits and questions

codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
giovanni-guidini: ah you are correct. Thanks.

codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
giovanni-guidini: So a "named" node (as far as I understand) is a node that has a type. Idk what nodes don't have a type tho.

The function docstring _is_ named, but its type is not "docstring", it's "expression".
So the function `_is_function_docstring` looks at other characteristics of that expression to determine if it's a function docstring or not.

codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
giovanni-guidini: "siblings" are nodes in the same "depth" in the code. So for example, global code and function definitions are all siblings (ordered by the line they are declared).
In the same way all lines in a function (in the same scope i.e. identation) are siblings. 
So if a node is the very first line of a function it won't have a previous sibling, because it's the first node of that scope.

Yes, it means that `current_line_number` doesn't have a sure ancestor in its scope (so we will look in the scope above)

codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
```diff
@@ -10,6 +10,43 @@ def visit(self, node):
         for c in node.children:
             self.visit(c)
 
+    def _is_function_docstring(self, node):
+        """Skips docstrings for funtions, such as this one.
+        Pytest doesn't include them in the report, so I don't think we should either,
```
scott-codecov, line 15: Is this specific to pytest or just coveragepy in general?

scott-codecov (approved):

codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py
giovanni-guidini: coveragepy in general

PR #271, opened 2023-09-29
fix: better handle errors in static analysis

giovanni-guidini:
This was needed already.
But ultimately inspired by a user tha had the following error:

```
File "codecov_cli/services/staticanalysis/analyzers/python/node_wrappers.py", line 41, in do_visit
    first_if_statement = first_if_statement.children[0]
IndexError: list index out of range
```

Despite my best efforts, I was only able to reproduce this error using invalid Python code
(i.e. an empty `if` statement)

Regardless of what caused this issue in the 1st place, there may be valid reasons
for someone to have malformed code in their repo (depending what the repo is 👀)
So we should be able to still upload _something_ and ignore the failures.

The tricky thing here is that without the static analysis data ATS can't work well.
And we can't be sure that the analyzer doesn't have bugs. I don't think this case in
particular is one of them, but I could be wrong.

So these changes actually are a way to surface processing errors AND ignore the files
that generated them, WHILE also surfacing errors to users. They are surfaced twice.
Ideally users would look at them and let us know if something is not right.

But maybe we should still fail the static analysis step in their CI? 🧐

### DIFF
--- a/codecov_cli/services/staticanalysis/__init__.py
+++ b/codecov_cli/services/staticanalysis/__init__.py
@@ -36,24 +36,15 @@ async def run_analysis_entrypoint(
 ):
     ff = select_file_finder(config)
     files = list(ff.find_files(folder, pattern, folders_to_exclude))
-    logger.info(f"Running the analyzer on {len(files)} files")
-    mapped_func = partial(analyze_file, config)
-    all_data = {}
-    file_metadata = []
-    with click.progressbar(
-        length=len(files),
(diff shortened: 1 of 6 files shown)


scott-codecov (approved):
Looks good to me.

Like we chatted about - it would make sense to me if we could catch this sort of malformed error a little higher up.  i.e. when tree-sitter is parsing the source into an AST does it know that an `if` with no children is not valid Python?

PR #272, opened 2023-09-29
Update --dry-run output

giovanni-guidini:
related to codecov/engineering-team#555

These changes update the output of `--dry-run` option so that:
1. Users can get a list of tests that are being skipped
2. Users have the option to parse the output from JSON

> ⚠️
>
>  The JSON output doesn't include the runner options in the tests to run list.
>  Instead there's the 'runner_options' key with that info.
>  This is to make sharding the list of tests easier

It's considerably more output, but given that we don't have clear guidelines on the
best way to do things I think it's better to provide all options.
Eventually adoption will converge to some way or another and we prioritize that.

### DIFF
--- a/codecov_cli/commands/labelanalysis.py
+++ b/codecov_cli/commands/labelanalysis.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import pathlib
 import time
@@ -55,13 +56,23 @@ logger = logging.getLogger("codecovcli")
 @click.option(
     "--dry-run",
     "dry_run",
-    help='Print list of tests to run and options that need to be added to the test runner as a space-separated list to stdout. Format is ATS_TESTS_TO_RUN="<options> <test_1> <test_2> ... <test_n>"',
+    help=(
+        "Print list of tests to run AND tests skipped (and options that need to be added to the test runner) to stdout. "
+        + "Also prints the same information in JSON format. "
+        + "JSON will have keys 'ats_tests_to_run', 'ats_tests_to_skip' and 'runner_options'. "
+        + "List of tests to run is prefixed with ATS_TESTS_TO_RUN= "
+        + "List of tests to skip is prefixed with ATS_TESTS_TO_SKIP="
+    ),
     is_flag=True,
 )
 @click.option(
     "--dry-run-output-path",
     "dry_run_output_path",
-    help="Prints the dry-run list into dry_run_output_path (in addition to stdout)",
+    help=(
+        "Prints the dry-run list (ATS_TESTS_TO_RUN) into dry_run_output_path (in addition to stdout)\n"
+        + "AND prints ATS_TESTS_TO_SKIP into dry_run_output_path_skipped\n"
+        + "AND prints dry-run JSON output into dry_run_output_path.json"
+    ),
     type=pathlib.Path,
     default=None,
 )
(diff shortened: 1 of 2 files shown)


matt-codecov (approved):
wait, so, we print a bunch of non-json info, and then we also print the json info? should that be a separate argument? like `--dry-run --format=json`? otherwise you have to copypaste or write logic to process the output

if this is still beta and we reserve the right to change the output format, won't block on that. but i think it's tidier to let the user say "just put the json (or envvar format) in stdout and i'll pipe it"

tests/commands/test_invoke_labelanalysis.py
```diff
@@ -332,20 +343,38 @@ def test_invoke_label_analysis_dry_run_with_output_path(
                     ],
                     obj={},
                 )
-                labels_file = Path("ats_output_path")
+                print(result)
+                print(result.output)
```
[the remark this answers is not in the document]
giovanni-guidini: The prints only show up if the test fails. In this case they are very useful to debug why the test failed. That's why some tests leave them there

giovanni-guidini:
> wait, so, we print a bunch of non-json info, and then we also print the json info? should that be a separate argument? like `--dry-run --format=json`? otherwise you have to copypaste or write logic to process the output
> 
> if this is still beta and we reserve the right to change the output format, won't block on that. but i think it's tidier to let the user say "just put the json (or envvar format) in stdout and i'll pipe it"

It is in beta and we do reserve the right to change the output format.
I agree that it's tidier that way, but we're still discovering which option is the best (which will eventually be the default one) and what this interface is going to be.

Right now the plan is "put all cards on the table".
In any case the output in stdout is dirty (meaning there are logs prior to the dry-run output) so there is some level of shell-fu required to clean it and use it anyway.

codecov_cli/commands/labelanalysis.py
```diff
@@ -55,13 +56,23 @@
 @click.option(
     "--dry-run",
     "dry_run",
-    help='Print list of tests to run and options that need to be added to the test runner as a space-separated list to stdout. Format is ATS_TESTS_TO_RUN="<options> <test_1> <test_2> ... <test_n>"',
+    help=(
+        "Print list of tests to run AND tests skipped (and options that need to be added to the test runner) to stdout. "
+        + "Also prints the same information in JSON format. "
+        + "JSON will have keys 'ats_tests_to_run', 'ats_tests_to_skip' and 'runner_options'. "
```
bukzor-sentryio, line 62: I think most users (on small-to-medium projects) will see "tests-to-skip" as unecessary noise.
I'd expect that it would be added to the output on a (default false) option; `--show-tests-skipped`, perhaps.

Entirely up to you. I'm happy either way.

codecov_cli/commands/labelanalysis.py
```diff
@@ -313,18 +324,50 @@ def _dry_run_output(
...
+        with open(str(dry_run_output_path) + ".json", "w") as fd:
+            fd.write(json.dumps(output_as_dict) + "\n")
+
+    click.echo(json.dumps(output_as_dict))
+    click.echo(
+        f"ATS_TESTS_TO_RUN={' '.join(runner.dry_run_runner_options + labels_run_wrapped_double_quotes)}"
+    )
+    click.echo(
+        f"ATS_TESTS_TO_SKIP={' '.join(runner.dry_run_runner_options + labels_skip_wrapped_double_quotes)}"
+    )
```
bukzor-sentryio, line 370: This stdout is nearly but not quite json parseable.
Most tools will do one of two things:
 
  1. print json on stdout, and non-json on stderr
  2. print a human-friendly form by default and only json if given --json
  
  If you choose one of these, the `--dry-run-output-path` option can go away: `tee(1)` fills that need, then.

codecov_cli/commands/labelanalysis.py
```diff
@@ -55,13 +56,23 @@
 @click.option(
     "--dry-run",
     "dry_run",
```
bukzor-sentryio, line 58: It's not clear (to me) what `label-analysis` does on a "wet" run. Add a function docstring or else pass `@command(help=...)` to help users understand.

bukzor-sentryio (commented):
Currently you have:

```python
if not dry_run:
    runner.process_labelanalysis_result(request_result)
else:
    _dry_run_output(...)
```
I recommend thinking about / modelling it a different way: categorizing the tests and running the tests are separate and equally important actions. Chaining list+run together can be the default, for simple use cases, but providing the ability to me to process the data between list and run is very powerful.

I'd model as three subcommands:

```
automated-test-selection -- based on your diff and past coverage, run the relevant tests
      categorize -- leverage automated test selection to generate test lists
      run -- run a automated-test-selection test list
```

The parent subcommand would be implemented as

```python
@click.command()
def automated_test_selection():
  testlist = ats.list()
  return ats.run()
```

tests/commands/test_invoke_labelanalysis.py
```diff
@@ -332,20 +343,38 @@ def test_invoke_label_analysis_dry_run_with_output_path(
...
                 with open(labels_file, "r") as fd:
                     assert fd.readlines() == [
-                        "--labels test_absent test_global test_in_diff\n"
+                        '--labels "test_absent" "test_global" "test_in_diff"\n'
+                    ]
+                with open(skip_labels_file, "r") as fd:
+                    assert fd.readlines() == ['--labels "test_present"\n']
+                with open(json_output, "r") as fd:
+                    assert fd.readlines() == [
+                        '{"runner_options": ["--labels"], "ats_tests_to_run": ["\\"test_absent\\"", "\\"test_global\\"", "\\"test_in_diff\\""], "ats_tests_to_skip": ["\\"test_present\\""]}\n'
```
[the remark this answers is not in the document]
giovanni-guidini: We recently had a customer run into trouble because one of their parametrized tests looked something like `path/to/test::TestClass::test_name[option -or another]`. When those were piped into `pytest` it assumed the `-or` was a command line option for pytest itself and failed.

That's why we are now double quoting all labels.

codecov_cli/commands/labelanalysis.py
giovanni-guidini: I agree with you. I think that will eventually be the case.

As I mentioned though, while the product is still in beta we are outputting everything to understand what is indeed considered "noise", and what usage patterns arise. Those datapoints will help us define better interfaces in the future.

giovanni-guidini:
> I recommend thinking about / modelling it a different way: categorizing the tests and running the tests are separate and equally important actions. Chaining list+run together can be the default, for simple use cases, but providing the ability to me to process the data between list and run is very powerful.

That is an interesting suggestion. We do want this tool be flexible and give a lot of power to the users on how to do things their way. The solution we came to was to bake that power and flexibility into it - the ability for you to process the data however you like.

This is done with [runners](https://docs.codecov.com/docs/label-analysis#runners). You can [create your own](https://docs.codecov.com/docs/label-analysis#custom-runners) and do whatever you want with the output of ATS. That is the reason that a runner is given the "raw output" of all lists.

I think it's important that we have this "plugin-like" structure for the adapter between ATS and the user's code if we want ATS to support many languages and testing tools.

codecov_cli/commands/labelanalysis.py
giovanni-guidini: I suppose you are right... possibly we should have not built it this way, but created a new runner to do what the `dry-run` option does. This would reinforce the intended way of using it

PR #273, opened 2023-10-02
chore: Remove requirement for pytest

giovanni-guidini:
pytest was required only by the `PythonStandardRunner`, and only in the `strict_mode`.
No one uses that.
At first it was develop because we were concerned that users would be picky by us running pytest with `subprocess.run`,
so `strict_mode` was an alternative. Since then people are starting to use ATS and that hasn't been an issue.

Plus with custom runners we can solve this problem without the complications of requiring pytest.
And there's a possibility (although I think it's very unlikely) that importing pytest from the CLI
might cause some of the issues we were seeing when running tests from it in ATS. Who knows.

In any case, removing pytest and all code associated with it.


joseph-sentry (commented):
can `_wait_pytest` and the imports for `Process` and `Queue` be removed form `python_standard_runner` be removed as well? Otherwise, lgtm!

joseph-sentry (dismissed):
other than my previous comment, lgtm!

joseph-sentry (approved):

PR #275, opened 2023-10-04
chore: better handle CancelledError from static analysis

giovanni-guidini:
context: codecov/engineering-team#643

Currently static analysis command is still very delicate and breaks easily.
Not so much (anymore) in the analysis part, but the uploading of files, specially in the first few runs, can cause indecipherable stack traces that simply state `CancelledError` was thrown.

This PR introduces some error handling for `send_single_upload_put` function and the general gathering of many such tasks in the hopes that we can fail gracefully (and not land on your face)

### DIFF
--- a/codecov_cli/services/staticanalysis/__init__.py
+++ b/codecov_cli/services/staticanalysis/__init__.py
@@ -109,7 +109,14 @@ async def run_analysis_entrypoint(
                 for el in files_that_need_upload:
                     all_tasks.append(send_single_upload_put(client, all_data, el))
                     bar.update(1, all_data[el["filepath"]])
-                resps = await asyncio.gather(*all_tasks)
+                try:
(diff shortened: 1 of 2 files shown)


codecov_cli/services/staticanalysis/__init__.py
```diff
@@ -188,35 +195,44 @@ async def process_files(
...
+                presigned_put, data=json.dumps(all_data[el["filepath"]])
+            )
+            if response.status_code < 300:
+                return {
+                    "status_code": response.status_code,
+                    "filepath": el["filepath"],
+                    "succeeded": True,
+                }
+            if response.status_code in retryable_statuses:
+                await asyncio.sleep(2**current_retry)
```
joseph-sentry, line 214: It'd be good to write a test that covers this line but otherwise lgtm

joseph-sentry (approved):

PR #276, opened 2023-10-05
disable file fixes

dana-yaish:
since node uploader supports turning file fixes off, we're doing it here too. 
https://github.com/codecov/engineering-team/issues/556
slack discussion for further context https://l.codecov.dev/RQNTLl


drazisil-codecov (approved):
Thank you!

PR #279, opened 2023-10-05
Adding builds for main and some reusable workflows

trent-codecov:


.github/workflows/create_release_pr.yml
```diff
@@ -9,6 +9,41 @@ on:
...
+          git_commit_gpgsign: true
+          git_config_global: true
+
+      - name: Create release branch
+        run: git checkout -b release/${{ github.event.inputs.versionName }}
+
+      - name: Update version and push
+        id: make-commit
+        run: |
+          sed -i 's/version="[0-9]\.[0-9]\.[0-9]"/version="${{ github.event.inputs.versionName }}"/g' setup.py
```
giovanni-guidini, line 33: Nice

.github/workflows/create_release_pr.yml
```diff
@@ -9,6 +9,41 @@ on:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Import GPG key
+        id: import-gpg
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.RELEASER_GPG_PRIVATE_KEY }}
+          git_user_signingkey: true
```
giovanni-guidini, line 23: Nice

.github/workflows/build_main.yml
```diff
@@ -0,0 +1,18 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Build
+
+on:
+  push:
+    branches:
+      - main
+      - feature/main-builds
```
giovanni-guidini, line 10: is this branch protected too? Would we want that?

giovanni-guidini (commented):
LGTM (but I've messed up this release process very much very recently so....:E)

One thing I don't understand is the publish to Pypi only happens if we manually trigger the build_for_pypi workflow?

.github/workflows/build_main.yml
trent-codecov: I need to remove this feature branch from that. I just had that for testing

.github/workflows/build_for_pypi.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          pip install -r requirements.txt
+          python setup.py build
+          python setup.py develop
+      - name: Build distributions for different platforms
+        run: |
+          pip install wheel
+          python setup.py sdist bdist_wheel --plat-name=manylinux2014_x86_64
+          python setup.py bdist_wheel --plat-name=macosx-12.6-x86_64
+          python setup.py bdist_wheel --plat-name=win_amd64
```
scott-codecov, line 35: Anthony recommended not using `bdist_wheel` and `--plat-name` (and instead using the `build` package instead): https://sentry.slack.com/archives/C04DL8SRBUZ/p1696528344746259?thread_ts=1696526939.075939&cid=C04DL8SRBUZ

Should we address that later?

scott-codecov (approved):

.github/workflows/build_for_pypi.yml
trent-codecov: Yes the goal was to take a pass at making this better as a different initiative. The goal here is to improve the overall workflow first.

PR #284, opened 2023-10-05
ref: fix version lookup and package build from source

asottile-sentry:
this is the minimum patch to fix building from source

tested with `python -m build`

I also added `python_requires` so that older pythons will get a useful error message if attempting to install an incompatible version


giovanni-guidini (dismissed):
Thank you for doing this 🥔🥔🥔

giovanni-guidini (approved):

PR #306, opened 2023-10-17
fix: Make static analysis handle a large number of uploads

giovanni-guidini:
This is more a fix around error handling and updating information to users than actual performance.

**Error Handling**

In terms of error handling, consider that the number of uploads is too big. We "start" all of them at the same time, but the number of connections is limited.

For uploads you stringify the data and try to grab a connection from the pool to upload. Then you need to _actually_ upload (e.g. write data to the pipe)

Some timeouts can exist in this process:
* `PoolTimeout` - waiting too long for a connection to become available
* `ReadTimeout` - waiting too long for a response back (these are the most important ones)

Given that we _will_ have thousands of files to upload it _will_ take a long time, but that's alright. So we can simply ignore those timeouts.

Then we need to balance the number of connections available. Through some empirical testing 100 is too much. If you have many big files trying to upload at the same time the COU time of each is too small, so it takes too long for the file to finish upload. This throws a 500 error from GCS.

So it's important that we don't try to upload too many files at the same time. With 20 connections I was able to upload ~6600 files with no errors.

**Update user info**

The progress bar was broken.
It was updating when the single upload was started, not when it finished. To fix that I just made it update when it finishes each upload.

Now we can properly see that it takes forever :D

### DIFF
--- a/codecov_cli/services/staticanalysis/__init__.py
+++ b/codecov_cli/services/staticanalysis/__init__.py
@@ -9,7 +9,6 @@ from pathlib import Path
 import click
 import httpx
 import requests
-import yaml
 
 from codecov_cli.helpers.config import CODECOV_API_URL
 from codecov_cli.services.staticanalysis.analyzers import get_best_analyzer
@@ -47,8 +46,16 @@ async def run_analysis_entrypoint(
     all_data = processing_results["all_data"]
     try:
         json_output = {"commit": commit, "filepaths": file_metadata}
+        logger.info(
+            "Sending files fingerprints to Codecov",
+            extra=dict(
+                extra_log_attributes=dict(
+                    files_effectively_analyzed=len(json_output["filepaths"])
+                )
+            ),
+        )
         logger.debug(
-            "Sending data for server",
+            "Data sent to Codecov",
             extra=dict(extra_log_attributes=dict(json_payload=json_output)),
         )
         upload_url = enterprise_url or CODECOV_API_URL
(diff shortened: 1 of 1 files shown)


giovanni-guidini:
If you are not convinced that `asyncio.as_completed` actually starts all tasks concurrently and returns them in order of completion - I wasn't because of the extra `await` inside the loop - please check this POC

```python
import asyncio
from datetime import datetime

async def task_fn(id, timeout, start_time):
    await asyncio.sleep(timeout / id)
    now = datetime.now()
    return f"{now} - {id} completed in {now - start_time}"

async def main():
    tasks = []
    for i in range(1, 11):
        tasks.append(task_fn(i, 10, datetime.now()))
    for task in asyncio.as_completed(tasks):
        r = await task
        print(r)

asyncio.run(main())
```

When running locally I get what I'd expect
```
$ python test_async.py 
2023-10-17 09:51:30.702246 - 10 completed in 0:00:01.001200
2023-10-17 09:51:30.813417 - 9 completed in 0:00:01.112372
2023-10-17 09:51:30.952330 - 8 completed in 0:00:01.251285
2023-10-17 09:51:31.131187 - 7 completed in 0:00:01.430143
2023-10-17 09:51:31.369108 - 6 completed in 0:00:01.668064
2023-10-17 09:51:31.702217 - 5 completed in 0:00:02.001174
2023-10-17 09:51:32.202236 - 4 completed in 0:00:02.501193
2023-10-17 09:51:33.035697 - 3 completed in 0:00:03.334655
2023-10-17 09:51:34.702088 - 2 completed in 0:00:05.001047
2023-10-17 09:51:39.701988 - 1 completed in 0:00:10.000954
```

codecov_cli/services/staticanalysis/__init__.py
```diff
@@ -96,36 +103,42 @@ async def run_analysis_entrypoint(
...
             length=len(files_that_need_upload),
-            label="Uploading files",
+            label=f"Upload info to storage",
         ) as bar:
-            async with httpx.AsyncClient() as client:
+            # It's better to have less files competing over CPU time when uploading
+            # Especially if we might have large files
+            limits = httpx.Limits(max_connections=20)
+            # Because there might be too many files to upload we will ignore most timeouts
+            timeout = httpx.Timeout(read=None, pool=None, connect=None, write=10.0)
```
scott-codecov, line 118: Not having any sort of read or connect timeout seems potentially not great but I guess in the worst case the task itself just times out.  That seems fine - we're not really blocking anything indefinitely.

scott-codecov (approved):

PR #308, opened 2023-10-17
Update build_assets.yml

joseph-sentry:


.github/workflows/build_assets.yml
```diff
@@ -61,6 +61,7 @@ jobs:
         run: |
           ${{matrix.CMD_REQS}}
           python setup.py build
+          pip install .
```
giovanni-guidini, line 64: What does this do?...

giovanni-guidini (approved):
Awesome 🥔🥔🥔

There's one line down there I don't quite understand, but LGTM

.github/workflows/build_assets.yml
thomasrockhu-codecov: It installs the CLI as the code in the repo (versus on pip or something else)

joseph-sentry: if `codecov-cli` is not installed the package metadata does not exist so we can't copy it over. By installing it, pyinstaller is able to find the metadata to copy over

giovanni-guidini: I see... thank you both for the explanations

giovanni-guidini (approved):

PR #310, opened 2023-10-18
Add test-file and disable-test-search options to CLI and start looking for testing result files

joseph-sentry:
This PR:
* adds the `test-file` and `disable-test-search` options to the CLI
* Creates the new `BaseFileFinder` and `TestingResultFileFinder` classes and modifies the `CoverageFileFinder` class
* Adds the testing_result_file_finder to the upload collection step and creates the `testing_result_files` field in `UploadCollectionResult`


codecov_cli/commands/upload.py
```diff
@@ -59,9 +59,24 @@ def _turn_env_vars_into_dict(ctx, params, value):
         multiple=True,
         default=[],
     ),
+    click.option(
+        "--test-file",
```
giovanni-guidini, line 63: I think you should hold on adding these options until we can actually process the data.
It's gonna be useless for users to have this if they can't use it, and possibly confusing as well

codecov_cli/commands/upload.py
```diff
@@ -250,6 +269,9 @@ def do_upload(
         coverage_files_search_explicitly_listed_files=list(
             coverage_files_search_explicitly_listed_files
         ),
+        test_result_files_search_explicitly_listed_files=list(
```
giovanni-guidini, line 272: We can't process the data, so this is effectively doing useless work as of now.
Let's turn it off until we can start ingesting this.

We should probably also feature flag it somehow. I'd start by setting the `disable_test_search` option default to `True`

giovanni-guidini (commented):
In general it looks good, but I left some comments

PR #315, opened 2023-10-19
chore: Fix helper text in label analysis

giovanni-guidini:
Since the previous change to label analysis the help text was outdated.
These changes fix it to be more accurate.

Also removing the test for the help text because it's more annoying than
it's helpful. It doesn't really _test_ anything, because the formatting
is done by click

### DIFF
--- a/codecov_cli/commands/labelanalysis.py
+++ b/codecov_cli/commands/labelanalysis.py
@@ -58,11 +58,8 @@ logger = logging.getLogger("codecovcli")
     "--dry-run",
     "dry_run",
     help=(
-        "Print list of tests to run AND tests skipped (and options that need to be added to the test runner) to stdout. "
-        + "Also prints the same information in JSON format. "
-        + "JSON will have keys 'ats_tests_to_run', 'ats_tests_to_skip' and 'runner_options'. "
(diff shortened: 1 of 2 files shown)


codecov_cli/commands/labelanalysis.py
```diff
@@ -57,18 +57,16 @@
     "--dry-run",
     "dry_run",
     help=(
-        "Print list of tests to run AND tests skipped (and options that need to be added to the test runner) to stdout. "
-        + "Also prints the same information in JSON format. "
-        + "JSON will have keys 'ats_tests_to_run', 'ats_tests_to_skip' and 'runner_options'. "
-        + "List of tests to run is prefixed with ATS_TESTS_TO_RUN= "
-        + "List of tests to skip is prefixed with ATS_TESTS_TO_SKIP="
+        "Print list of tests to run AND tests skipped AND options that need to be added to the test runner to stdout. "
+        + "Choose format with --dry-run-format option. Default is JSON. "
```
thomasrockhu-codecov, line 61: can we list the other options?

giovanni-guidini: It does list them in the helper text of `--dry-run-format`

scott-codecov (approved):

PR #316, opened 2023-10-20
Allow JWT tokens

zetaab:
As https://github.com/codecov/codecov-api/pull/177 is merged, the token format can be either JWT or UUID.

This PR is blocked for another PR https://github.com/codecov/codecov-action/pull/1054


zetaab:
@scott-codecov that one ci test is failing. However, at least I tried to remove all UUID checks. Is that CI job somehow using old CLI or how it works? Can that be skipped or what is the situation?

zetaab:
blah there is more, I will remove those

zetaab:
yeah looks like its loading current 0.4.0 version of CLI and trying tests against that. However, it cannot work because the uuid logic is removed in this PR.

PR #318, opened 2023-10-23
Add TestRun, TestRunGroup and JUnit parser classes

joseph-sentry:
This PR:
* Creates the Test and Testrun classes to have a common representation of testing results between all parsers
* Creates JUnitXMLParser class to be able to parse junit.xml files

### DIFF
--- a/codecov_cli/parsers/base.py
+++ b/codecov_cli/parsers/base.py
@@ -1,60 +1,44 @@
+from dataclasses import dataclass
 from datetime import datetime, timedelta
-from typing import List
+from json import JSONEncoder
+from typing import Any, List
 
 
 class ParsingError(Exception):
     ...
 
 
-class Test(object):
-    def __init__(self, name: str, status: bool, duration: timedelta):
-        self.name = name
-        self.status = status
-        self.duration = duration
+class ParserJSONEncoder(JSONEncoder):
+    def default(self, o: Any) -> Any:
+        if isinstance(o, TestRunGroup | TestRun):
+            return o.__dict__
+        elif isinstance(o, timedelta):
+            return o.total_seconds()
+        elif isinstance(o, datetime):
+            return o.isoformat()
+        else:
+            raise ParsingError("Error encoding object to JSON")
 
-    def __repr__(self):
-        return f"{self.name}:{self.status}:{self.duration.total_seconds()}::"
+
+@dataclass
+class TestRun(object):
+    name: str
+    status: bool
+    duration: timedelta
 
 
-class Testrun(object):
-    def __init__(
-        self,
-        name: str,
-        timestamp: datetime,
-        time: timedelta,
-        testcases: List[Test],
-        failures: int,
-        errors: int,
-        skipped: int,
-        total: int,
-    ):
-        self.testcases = testcases or []
-        self.name = name
-        self.time = time
-        self.timestamp = timestamp
-        self.failures = failures
-        self.errors = errors
-        self.skipped = skipped
-        self.total = total
-
-    def to_str(self):
-        return (
-            ";".join(
-                [
-                    self.name,
-                    str(self.timestamp.timestamp()),
-                    str(self.time.total_seconds()),
-                    str(self.failures),
-                    str(self.errors),
-                    str(self.skipped),
-                    str(self.total),
-                    str(self.testcases),
-                ]
-            )
-            + ";"
-        )
+@dataclass
+class TestRunGroup(object):
+    name: str
+    timestamp: datetime
+    time: timedelta
+    testruns: List[TestRun]
+    failures: int
+    errors: int
+    skipped: int
+    total: int
 
 
 class Parser:
-    def parse(self, file_content) -> List[Testrun]:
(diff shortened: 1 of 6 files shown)


codecov_cli/parsers/base.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class ParsingError(Exception):
+    ...
+
+
+class Test(object):
+    def __init__(self, name: str, status: bool, duration: timedelta):
+        self.name = name
+        self.status = status
```
giovanni-guidini, line 12: I think `status` should be an Enum rather than a bool.
Or change it to "passed" or "successful"

codecov_cli/parsers/base.py
```diff
@@ -0,0 +1,60 @@
...
+        name: str,
+        timestamp: datetime,
+        time: timedelta,
+        testcases: List[Test],
+        failures: int,
+        errors: int,
+        skipped: int,
+        total: int,
+    ):
+        self.testcases = testcases or []
```
giovanni-guidini, line 31: If you don't specify a default value for test cases (e.g. `testcases: List[Test] = None` then it can't ever be `None`, right?

(if it can then the type hint should be `Optional`)

codecov_cli/parsers/junit.py
```diff
@@ -0,0 +1,51 @@
...
+        testsuites = [
+            self._create_testsuite(testsuite_xml)
+            for testsuite_xml in processed.iter("testsuite")
+        ]
+
+        return testsuites
+
+    def _create_testcase(self, testcase_xml: etree.Element):
+        return Test(
+            f"{testcase_xml.get('classname')}.{testcase_xml.get('name')}",
```
giovanni-guidini, line 30: Can this be transformed in the format that we use for labels?
That is `path/to/test.py::TestClassIfAny::test_function_name`.

It would be helpful to use this in an ATS setting to have the test names follow this format.

codecov_cli/parsers/junit.py
```diff
@@ -0,0 +1,51 @@
...
+class JUnitXMLParser(Parser):
+    def __init__(self):
+        self._parser = etree.XMLParser(recover=True, resolve_entities=False)
+
+    def parse(self, file_content) -> List[Testrun]:
+        processed = self._parse_xml(file_content)
+        if processed is None or len(processed) == 0:
+            raise ParsingError("Error parsing XML file")
+
+        testsuites = [
```
giovanni-guidini, line 21: You are calling this `testsuites` but the type name is `TestRun`. I think you should pick one and stick to it. Whatever makes the most sense in the context.

codecov_cli/parsers/base.py
```diff
@@ -0,0 +1,60 @@
...
+        self.name = name
+        self.time = time
+        self.timestamp = timestamp
+        self.failures = failures
+        self.errors = errors
+        self.skipped = skipped
+        self.total = total
+
+    def to_str(self):
```
giovanni-guidini, line 40: The uploader uploads JSON data... maybe we should have the encoding/decoding be to JSON?

If this is for debug purposes (e.g. printing the class nicely into console) than I believe the right function to override is `__repr__`.

codecov_cli/parsers/base.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class Test(object):
+    def __init__(self, name: str, status: bool, duration: timedelta):
+        self.name = name
+        self.status = status
+        self.duration = duration
+
+    def __repr__(self):
+        return f"{self.name}:{self.status}:{self.duration.total_seconds()}::"
```
giovanni-guidini, line 16: Is this to be used as the encoding/decoding or just debugging?
If just for debugging than it's very confusing.

If it's for the encoding/decoding I think it's ok to use JSON directly.
I understand that the payload will be larger, but we compress it before uploading. And making the data easier to understand and work with seems more relevant... I could be wrong, if you have data backing up the fact that unless we encode the data in this compact way the resulting payload will be huge.

joseph-sentry:
> Can this be transformed in the format that we use for labels? That is `path/to/test.py::TestClassIfAny::test_function_name`.
> 

It would be helpful to use this in an ATS setting to have the test names follow this format.
So after having done some research the answer to this question is not straightforward. It turns out that the pytest junitxml plugin can generate two different versions of junit report, "xunit1" and "xunit2", the "xunit1" version includes a file with each testcase that is generated in the report which should enable this, but I have not implemented support for this in this PR. "xunit2" for which you can find the schema [here](https://github.com/jenkinsci/xunit-plugin/blob/xunit-2.3.2/src/main/resources/org/jenkinsci/plugins/xunit/types/model/xsd/junit-10.xsd) can support including the file in the `testsuite` xml eleement, but at the moment it doesn't look like the pytest plugin generates that tag, so it won't be easy to create an ATS formatted name in this case, we would have to try and match the `classname + name` to the label, this is made easier by the fact that the path to the file in included in the classname (see [this](https://github.com/pytest-dev/pytest/blob/main/src/_pytest/junitxml.py#L453-L462))

Issue #320, opened 2023-10-26
UnboundLocalError: local variable 'lineno' referenced before assignment [18] Failed to execute script 'main' due to unhandled exception!


rj-codecov:
A customer is trying to run an upload via the CLI and is getting this error. Joseph Sawaya has determined that this may be caused by trying to apply fixes (report fixes) to an empty file, which is the only way lineno would be unset.

PR #327, opened 2023-11-06
Fix lineno being unset

joseph-sentry:
If lineno is unset that means that the file is empty thus the eof should be 0, so lineno will be -1.

Fixes: https://github.com/codecov/codecov-cli/issues/320

### DIFF
--- a/codecov_cli/services/upload/upload_collector.py
+++ b/codecov_cli/services/upload/upload_collector.py
@@ -112,6 +112,9 @@ class UploadCollector(object):
 
         try:
             with open(filename, "r") as f:
+                # If lineno is unset that means that the
+                # file is empty thus the eof should be 0
+                # so lineno will be set to -1 here
                 lineno = -1
                 for lineno, line_content in enumerate(f):
                     if any(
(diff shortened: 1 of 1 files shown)


codecov_cli/services/upload/upload_collector.py
```diff
@@ -112,6 +112,7 @@ def _get_file_fixes(
 
         try:
             with open(filename, "r") as f:
+                lineno = -1
```
giovanni-guidini, line 115: Maybe a comment here to explain why it starts at -1 and not 0?

giovanni-guidini (dismissed):

giovanni-guidini (approved):

codecov_cli/services/upload/upload_collector.py
matt-codecov: not sure this actually does anything. `lineno` will be overwritten by whatever `enumerate()` spits out on the next line, right?

is `enumerate(f, start=-1)` what you want?

matt-codecov (changes requested):

codecov_cli/services/upload/upload_collector.py
matt-codecov: oh i see, we set `eof` outside the loop. could you add a comment before the for loop saying, like, "if not empty, we overwrite lineno. if empty, it stays -1" so nobody deletes your line in the future bc they think it's extraneous

matt-codecov (approved):

giovanni-guidini (approved):

PR #328, opened 2023-11-07
check if the PR is a fork PR

dana-yaish:
This is the first PR for tokenless upload and branch conflation bug fix where we check if the PR is a fork PR. 
I'm not using any auth tokens as forked PRs are public for open source repos. And a forked repo made from a public upstream cannot change its visibility to private. 
This PR implements getting PR info from github only, other providers are needed still, but pushing my code to get initial review 
We'll need to add the fork PR check in each command we want to support tokenless for. For uploads we'll need to add it to create_commit, create_report and do_upload commands logic. But, create_report does not have reference to the PR number, which means we'll need to add it there. 

https://github.com/codecov/engineering-team/issues/733

### DIFF
--- a/codecov_cli/helpers/git.py
+++ b/codecov_cli/helpers/git.py
@@ -3,6 +3,9 @@ import re
 from enum import Enum
 from urllib.parse import urlparse
 
+from codecov_cli.helpers.encoder import decode_slug
+from codecov_cli.helpers.git_services.github import Github
+
 slug_regex = re.compile(r"[^/\s]+\/[^/\s]+$")
 
 logger = logging.getLogger("codecovcli")
@@ -15,6 +18,11 @@ class GitService(Enum):
     GITHUB_ENTERPRISE = "github_enterprise"
     GITLAB_ENTERPRISE = "gitlab_enterprise"
     BITBUCKET_SERVER = "bitbucket_server"
+
+
+def get_git_service(git):
+    if git == "github":
+        return Github()
 
 
 def parse_slug(remote_repo_url: str):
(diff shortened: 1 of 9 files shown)


codecov_cli/helpers/git.py
```diff
@@ -82,3 +90,13 @@ def parse_git_service(remote_repo_url: str):
...
         )
         return None
+
+
+def is_fork_pr(pr_num, slug, service):
+    decoded_slug = decode_slug(slug)
+    git_service = get_git_service(service)
+    if git_service:
+        pull_dict = git_service.get_pull_request(decoded_slug, pr_num)
+        if pull_dict and pull_dict["head"]["slug"] != decoded_slug:
```
matt-codecov, line 100: do we need to decode the slug? if the PR API request gives us both the head and the base, we can just see if they're different?
```
if pull_dict and pull_dict["head"]["slug"] != pull_dict["base"]["slug"]:
```

codecov_cli/helpers/git_services/github.py
```diff
@@ -0,0 +1,32 @@
+import json
+
+import requests
+
+
+class Github:
+    api_url = "https://api.github.com"
+    api_version = "2022-11-28"
+
+    def get_pull_request(self, slug, pr_number):
```
matt-codecov, line 10: i think this is tested indirectly in `test_git.py` but it'd be nice if it had its own tests

matt-codecov (approved):
lgtm, approving to spare you timezone hell but pls read the comments and decide how to address them

codecov_cli/helpers/git.py
dana-yaish: you're right we ca do it both ways, but we need to decode it since we need it in the API request to get the pull request

codecov_cli/helpers/git_services/github.py
dana-yaish: I agree, added tests

matt-codecov (approved):
thanks for adding the test! other changes also look good

PR #330, opened 2023-11-08
Change thread pool method in process_files

joseph-sentry:


codecov_cli/services/staticanalysis/__init__.py
```diff
@@ -183,11 +184,15 @@ async def process_files(
     all_data = {}
     file_metadata = []
     errors = {}
+    if sys.platform.startswith("win32") or sys.platform.startswith("darwin"):
```
giovanni-guidini, line 187: Add a comment with link to https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods please

giovanni-guidini (approved):
Looks good. I'd love to see a comment there.
Looking at this again I realized that we are simply using the defaults tho, right? I wonder if we should just remove `get_context` from there and let the lib do its thing

giovanni-guidini (approved):

PR #331, opened 2023-11-10
amend branch names to be forked-slug:branch-name

dana-yaish:
- taking get_pull out of is_fork_pr for usability. I don't think it's the responsibility of is_fork_pr to do the request and get the pull's information 
- overriding branch in commit's endpoint since that's the only command that actually uses it and sends it in part of the request.

### DIFF
--- a/codecov_cli/helpers/git.py
+++ b/codecov_cli/helpers/git.py
@@ -97,11 +97,7 @@ def is_fork_pr(pull_dict):
     takes in dict: pull_dict
     returns true if PR is made in a fork context, false if not.
     """
-    return (
-        True
-        if pull_dict and pull_dict["head"]["slug"] != pull_dict["base"]["slug"]
-        else False
-    )
+    return pull_dict and pull_dict["head"]["slug"] != pull_dict["base"]["slug"]
 
 
 def get_pull(service, slug, pr_num):
--- a/codecov_cli/helpers/git.py
+++ b/codecov_cli/helpers/git.py
@@ -92,14 +92,24 @@ def parse_git_service(remote_repo_url: str):
         return None
 
 
-def is_fork_pr(pr_num, slug, service):
+def is_fork_pr(pull_dict):
     """
-    takes in pull request number, slug in the owner/repo format, and the git service e.g. github, gitlab etc.
+    takes in dict: pull_dict
     returns true if PR is made in a fork context, false if not.
+    """
+    return (
+        True
+        if pull_dict and pull_dict["head"]["slug"] != pull_dict["base"]["slug"]
+        else False
+    )
+
+
+def get_pull(service, slug, pr_num):
+    """
+    takes in str git service e.g. github, gitlab etc., slug in the owner/repo format, and the pull request number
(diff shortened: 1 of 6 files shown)


codecov_cli/helpers/git.py
```diff
@@ -92,14 +92,24 @@ def parse_git_service(remote_repo_url: str):
...
     """
-    takes in pull request number, slug in the owner/repo format, and the git service e.g. github, gitlab etc.
+    takes in dict: pull_dict
     returns true if PR is made in a fork context, false if not.
     """
+    return (
+        True
+        if pull_dict and pull_dict["head"]["slug"] != pull_dict["base"]["slug"]
+        else False
+    )
```
matt-codecov, line 104: rather than use a ternary here you could do this in one line like:
```
return pull_dict and pull_dict["head"]["slug"] != pull_dict["base"]["slug"]
```

codecov_cli/services/commit/__init__.py
```diff
@@ -44,18 +44,22 @@ def create_commit_logic(
 def send_commit_data(
     commit_sha, parent_sha, pr, branch, slug, token, service, enterprise_url
 ):
+    decoded_slug = decode_slug(slug)
+    pull_dict = get_pull(service, decoded_slug, pr) if not token else None
+    if is_fork_pr(pull_dict):
+        headers = {}
+        branch = pull_dict["head"]["slug"] + ":" + branch
```
matt-codecov, line 51: i think this is good here, just a note for later:

since `:` is a legal character in real branch names, i think in the API when we double-check that this PR is from a fork we should set a different field entirely in the database that the UI can check instead of trying to parse it out of the branch name (see ugly screenshot mockup)

Screenshot 2023-11-13 at 10 40 40 AM

matt-codecov (approved):

codecov_cli/services/commit/__init__.py
dana-yaish: are you sure it's a valid character? I tried to create a branch with a ":" but says it's invalid, and the documentations say that too. see [here](https://git-scm.com/docs/git-check-ref-format) and [here](https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names#about-branch-and-tag-names)

codecov_cli/helpers/git.py
dana-yaish: yup missed that. Thank you I'll update it

codecov_cli/services/commit/__init__.py
dana-yaish: but would be a nice UI addition for PRs coming from a fork

matt-codecov: oh i stand corrected, you're right! i should have checked official docs, i just read some page that said git branch names were "very permissive"

matt-codecov (approved):
still lgtm :)

does the repo automatically dismiss reviews or are you doing that manually?

dana-yaish:
@matt-codecov nope it's automatic

PR #334, opened 2023-12-05
add X-Tokenless header when uploading from fork

giovanni-guidini:
Public forks will accept tokenless uploads.
Currently we were just sending an empty header (no Authorization).
These changes add a header `X-Tokenless: fork_slug` so we know
easily that the request is from a fork, and which fork it's from.

I also have a tendency to compulsively add typehints to complex
types.

### Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. In 2022 this entity acquired Codecov and as result Sentry is going to need some rights from me in order to utilize my contributions in this PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.


giovanni-guidini:
Ops messed up the branches. Sorry 😅

PR #335, opened 2023-12-05
add X-Tokenless header when uploading from fork

giovanni-guidini:
Public forks will accept tokenless uploads.
Currently we were just sending an empty header (no Authorization). These changes add a header `X-Tokenless: fork_slug` so we know easily that the request is from a fork, and which fork it's from.

I also have a tendency to compulsively add typehints to complex types.

### DIFF
--- a/codecov_cli/helpers/git.py
+++ b/codecov_cli/helpers/git.py
@@ -1,9 +1,11 @@
 import logging
 import re
 from enum import Enum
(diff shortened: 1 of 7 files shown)


scott-codecov (approved):
> I also have a tendency to compulsively add typehints to complex types.

Very helpful, thanks!

PR #342, opened 2023-12-13
feat: add X-Tokenless-PR header

giovanni-guidini:
If you look at https://github.com/codecov/codecov-api/pull/304 you'll notice that it expects 2 headers - `X-Tokenless` and `X-Tokenless-PR`.

We were not yet sending `X-Tokenless-PR` (because I only realized we'd need it after doing the API side of things).

These changes add that header to tokenless requests.

ps: also sneaking in a change to `.gitignore`.


.gitignore
```diff
@@ -163,3 +163,5 @@ cython_debug/
 # Vim temporary files
 *.swp
 *.swo
+
+.debug
```
adrian-codecov, line 167: Sneaky

adrian-codecov (dismissed):
Lgtm but we're missing coverage on 2 lines it seems, can you check that out pls?

adrian-codecov (approved):

PR #346, opened 2024-01-02
feat+fix: preserve test collection order

giovanni-guidini:
We run into issues with test ordering with some frequency in the ATS
step in CI. This is because ATS scrambles the test order and this
uncovers certain dependencies within tests.

Sure there's some value in doing that, but fixing these issues is
somewhat annoying and actually disrupts the workflow more than is
helpful. Scrambling test order needs to be deliberate, not accidental

These changes make sure that we preserve the order that pytest
collected tests when deciding the order to run them.

Notice that it can still run into ordering issues if a test that
depends on another test does't get picked to run.

### DIFF
--- a/codecov_cli/commands/labelanalysis.py
+++ b/codecov_cli/commands/labelanalysis.py
@@ -192,7 +192,7 @@ def label_analysis(
                 runner.process_labelanalysis_result(request_result)
             else:
                 _dry_run_output(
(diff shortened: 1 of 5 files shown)


OKEAMAH (commented):
Okeamah.eth

giovanni-guidini:
@thomasrockhu-codecov I think it's pretty safe to close this. We no longer need it.

PR #362, opened 2024-01-26
feat: add force flag to empty upload

joseph-sentry:
Fixes: https://github.com/codecov/engineering-team/issues/1016
API change: https://github.com/codecov/codecov-api/pull/362

### DIFF
--- a/codecov_cli/commands/empty_upload.py
+++ b/codecov_cli/commands/empty_upload.py
@@ -12,7 +12,7 @@ logger = logging.getLogger("codecovcli")
 
 
 @click.command()
-@click.option("--force/--no-force", default=False)
+@click.option("--force", is_flag=True, default=False)
 @global_options
 @click.pass_context
 def empty_upload(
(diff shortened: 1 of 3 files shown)


giovanni-guidini (approved):
LGTM

codecov_cli/commands/empty_upload.py
```diff
@@ -12,11 +12,13 @@
 
 
 @click.command()
+@click.option("--force/--no-force", default=False)
```
giovanni-guidini, line 15: I'll leave the comment here but know that this is a _personal_ opinion and you can ignore this. It's also veeery nit. But if it resonates with you... :P 

I saw that this pattern is also used in `label_analysis` command. But I don't particularly like it. Having a `--no-force` flag makes no sense cause it's the default anyway. And I don't like the syntax either.

I'd recommend `@click.option("--force", is_flag=True, default=False)` instead

giovanni-guidini (approved):

PR #363, opened 2024-01-29
build: upload test results

joseph-sentry:
Uploads test results to codecov

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,11 +76,12 @@ jobs:
         pip install -r tests/requirements.txt
     - name: Test with pytest
       run: |
-        pytest --cov
+        pytest --cov --junitxml=junit.xml
     - name: Dogfooding codecov-cli
       if: ${{ !github.event.pull_request.head.repo.fork && github.repository_owner == 'codecov' }}
       run: |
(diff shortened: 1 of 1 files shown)


adrian-codecov (dismissed):

.github/workflows/ci.yml
```diff
@@ -76,11 +76,12 @@ jobs:
         pip install -r tests/requirements.txt
     - name: Test with pytest
       run: |
-        pytest --cov
+        pytest --cov --junitxml=junit.xml
     - name: Dogfooding codecov-cli
       if: ${{ !github.event.pull_request.head.repo.fork && github.repository_owner == 'codecov' }}
       run: |
         codecovcli do-upload --fail-on-error -t ${{ secrets.CODECOV_TOKEN }} --plugin pycoverage --flag python${{matrix.python-version}}
```
giovanni-guidini, line 83: This might be a stupid question, but for my own understanding, do we expect that the regular upload will also include the test result file (and won't find a parser for it and then will fail) or it's not even picked up?

giovanni-guidini (approved):
Excited to see this in ours repos! o/

.github/workflows/ci.yml
joseph-sentry: it might add it to the coverage uploads but there should not be a parser for it.

Currrent settings are that the coverage file finder will not ignore files if they have "junit" in the name [source](https://github.com/codecov/codecov-cli/blob/main/codecov_cli/services/upload/file_finder.py#L11-L177)

Issue #365, opened 2024-02-01
CLI does not detect codecov.yml files from allowed locations


thomasrockhu-codecov:
The codecov.yml file should be read from

`/.github`
`/dev`
`/`

directories as

`codecov.yml`
`codecov.yaml`
`.codecov.yml`
`.codecov.yaml`

thomasrockhu-codecov:
Will get confirmation after deploy

thomasrockhu-codecov:
Shipped in `0.4.6`

PR #368, opened 2024-02-02
fix: search acceptable paths for codecov yaml if not passed in

matt-codecov:
fixes https://github.com/codecov/codecov-cli/issues/365

the node uploader [searched for a codecov yaml in one of 12 paths](https://github.com/codecov/uploader/blob/1e505ba8a27a1e3311a7a8db7f314e3705a2ee08/src/helpers/token.ts#L69-L76). this PR searches the same 12 paths when looking for the codecov yaml on startup

### DIFF
--- a/codecov_cli/helpers/config.py
+++ b/codecov_cli/helpers/config.py
@@ -1,20 +1,65 @@
 import logging
 import pathlib
+import typing
 
 import yaml
+
+from codecov_cli.helpers.versioning_systems import get_versioning_system
 
 logger = logging.getLogger("codecovcli")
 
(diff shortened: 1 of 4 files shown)


giovanni-guidini (approved):
LGTM

I don't really like all the empty `codecov.ymls` in the project though. I'd suggest using [tempfile](https://github.com/codecov/codecov-cli/blob/main/tests/services/upload/test_coverage_file_finder.py#L1) to create a fake project structure on-the-fly
(and use `pytest.mark.parametrize`)

But you know, small suggestion that should not prevent this from being merged.

Issue #370, opened 2024-02-02
Undefined method in error handling


MalteHerrmann:
Since updating to the CI action v4, there is the following error output, when receiving a `500` response:

```
error - 2024-02-02 10:31:59,259 -- Commit creating failed: {"error": "Server Error (500)"}
Traceback (most recent call last):
  File "codecov_cli/main.py", line 81, in <module>
  File "codecov_cli/main.py", line 77, in run
  File "click/core.py", line 1157, in __call__
  File "click/core.py", line 1078, in main
  File "click/core.py", line 1688, in invoke
  File "click/core.py", line 14[34](https://github.com/evmos/evmos/actions/runs/7754399543/job/21148217127?pr=2317#step:6:35), in invoke
  File "click/core.py", line 783, in invoke
  File "click/decorators.py", line 33, in new_func
  File "codecov_cli/commands/commit.py", line 64, in create_commit
  File "codecov_cli/services/commit/__init__.py", line [39](https://github.com/evmos/evmos/actions/runs/7754399543/job/21148217127?pr=2317#step:6:40), in create_commit_logic
  File "codecov_cli/helpers/request.py", line 133, in log_warnings_and_errors_if_any
NameError: name 'exit' is not defined
[32155] Failed to execute script 'main' due to unhandled exception!
Error: Codecov: Failed to properly create commit: The process '/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov' failed with exit code 1
```

The reason for this is that the `exit` method used in `helpers/request.py` is not defined/imported.

https://github.com/codecov/codecov-cli/blob/v0.4.6/codecov_cli/helpers/request.py#L130-L133

thomasrockhu-codecov:
@joseph-sentry this has been deployed in latest?

PR #371, opened 2024-02-02
fix: use sys.exit() instead of just exit()

joseph-sentry:
related to: https://github.com/codecov/codecov-cli/issues/370

### DIFF
--- a/codecov_cli/helpers/request.py
+++ b/codecov_cli/helpers/request.py
@@ -1,4 +1,5 @@
 import logging
+from sys import exit
 from time import sleep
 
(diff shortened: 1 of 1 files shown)


joseph-sentry:
since the error message in the original ticket mentions the `exit` command not being defined I'm assuming it's because we're using the `exit` function not in `sys` which apparently is meant only for use in the python REPL and shouldn't be used in programs, in the case that site is not loaded (see: https://docs.python.org/3/library/constants.html#constants-added-by-the-site-module)

PR #373, opened 2024-02-02
feat: make noop plugin recognizable

giovanni-guidini:
context: https://github.com/codecov/feedback/issues/258

Currently there's no way to disable plugins.
We don't want to change the default to preserve compatibility with the uploader little magic tricks that were executed by default (as far as I know).

These changes allow you to pass the `--plugin noop` option to the `create-upload` command in order to not execute any plugin.

### DIFF
--- a/tests/plugins/test_instantiation.py
+++ b/tests/plugins/test_instantiation.py
@@ -115,7 +115,7 @@ def test_get_plugin_xcode():
     assert isinstance(res, XcodePlugin)
 
 
-def test_get_plugin_xcode():
+def test_get_plugin_noop():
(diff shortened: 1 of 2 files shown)


tests/plugins/test_instantiation.py
```diff
@@ -115,6 +115,11 @@ def test_get_plugin_xcode():
     assert isinstance(res, XcodePlugin)
 
 
+def test_get_plugin_xcode():
```
joseph-sentry, line 118: rename to test_get_plugin_noop

joseph-sentry (approved):

matt-codecov (approved):

Issue #372, opened 2024-02-02
Correct SHA may not being chosen when used in a pre-merge commit


drazisil-codecov:
Issue: My notifications are not working when using the CLI.

In the case that the incorrect SHA is picked, Codecov will show the commit message as `merge <SHA>`

Workaround: You can manually pass the correct SHA to the CLI with the `--sha` flag.

If you are using the CircleCI orb, or the GitHub action, please see that wrappers's docs for the method to manually pass the SHA.

We are aware of this issue.

drazisil-codecov:
The Uploader uses the following logic to detect the correct SHA in a pre-merge commit:

https://github.com/codecov/uploader/blob/1e505ba8a27a1e3311a7a8db7f314e3705a2ee08/src/ci_providers/provider_githubactions.ts#L97-L122

rohan-at-sentry:
@drazisil-codecov can you check if the following are related

1. https://github.com/codecov/feedback/issues/126#issuecomment-1931862637
2. https://github.com/codecov/codecov-action/issues/1274#issuecomment-1934469452

PR #375, opened 2024-02-06
fix: update merge commit command

thomasrockhu-codecov:
fixes https://github.com/codecov/codecov-cli/issues/372

This PR mirrors what is done in the [uploader](https://github.com/codecov/uploader/blob/main/src/ci_providers/provider_githubactions.ts#L104-L119) and [bash uploader](https://github.com/codecov/codecov-bash/blob/master/codecov#L865-L874) (what a throwback) to collect the proper commit for merge commits.

### DIFF
--- a/codecov_cli/helpers/ci_adapters/github_actions.py
+++ b/codecov_cli/helpers/ci_adapters/github_actions.py
@@ -37,7 +37,7 @@ class GithubActionsCIAdapter(CIAdapterBase):
                 logger.info(
                     "->  Issue detecting commit SHA. Please run actions/checkout with fetch-depth > 1 or set to 0"
                 )
-        except TypeError:  # For the re.match
+        except AttributeError, TypeError:  # For the re.match and .decode
             logger.info(
                 f"    Commit with SHA {commit} of PR {pr} is not a merge commit"
             )
--- a/codecov_cli/helpers/ci_adapters/github_actions.py
+++ b/codecov_cli/helpers/ci_adapters/github_actions.py
@@ -1,8 +1,11 @@
+import logging
 import os
 import re
 import subprocess
 
 from codecov_cli.helpers.ci_adapters.base import CIAdapterBase
+
+logger = logging.getLogger("codecovcli")
 
 
 class GithubActionsCIAdapter(CIAdapterBase):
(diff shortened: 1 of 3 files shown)


codecov_cli/helpers/ci_adapters/github_actions.py
```diff
@@ -17,14 +20,27 @@ def _get_commit_sha(self):
...
-        completed_subprocess = subprocess.run(
-            ["git", "rev-parse", "HEAD^@"], capture_output=True
-        )
-
-        parents_hash = completed_subprocess.stdout.decode().strip().splitlines()
-        if len(parents_hash) == 2:
-            return parents_hash[1]
+        merge_commit_regex = r"^[a-z0-9]{40} [a-z0-9]{40}$"
+        merge_commit_message = subprocess.run(
+            ["git", "show", "--no-patch", "--format=%P"],
```
giovanni-guidini, line 25: You don't actually have to change this adapter's logic. The command being used accounts for this fact and already "fixes" the merge commit if that's the case.

```
$ git rev-parse HEAD^@
1b00e154fb8a8a7ad09da4f9c6c95e70c5fa1565
211b95514359270c24c3e4e2058b3eef83db75e4 <-- this would be returned by the old logic
$ git show --no-patch --format=%P | cat
1b00e154fb8a8a7ad09da4f9c6c95e70c5fa1565 211b95514359270c24c3e4e2058b3eef83db75e4

giovanni-guidini (changes requested):
The fix seems fine, but applying it only to GHActions adapter will not actually solve the problem described in #372 .

That problem affects mostly _unknown_ CI providers (that are considered "local").

I suggest moving the fix logic to a helper function, possibly in `codecov_cli.helpers.git.py` and apply this behavior to all CI adapters by possibly correcting the commit suggested [here](https://github.com/codecov/codecov-cli/blob/da50de0d1dc9f33d01c616dcafd76a0619a0af3c/codecov_cli/fallbacks.py#L32) 

There doesn't seem to be a need to change the GHActions adapter at this time.

giovanni-guidini:
I'm closing this PR. I moved his logic to https://github.com/codecov/codecov-cli/pull/397 instead (but using the commend from github actions)

It aims to be the same thing, but in a different part of the code.

PR #376, opened 2024-02-07
Define git_service config getter for Github actions

kdarkhan:
This is my attempt at fixing the issue https://github.com/google/mdbook-i18n-helpers/issues/173.

I am not very familiar with this codebase but I think it might fix the issue.

My Github workflows implements an unusual technique for codecov uploads which I described in [this thread](https://github.com/codecov/feedback/issues/126#issuecomment-1924413592).

It used to work in `codecov-action` v3 and stopped working in v4. The error I am getting is `Commit creating failed: ["Service not found: none"]`.

I suspect this is due to missing implementation of [`_get_git_service`](https://github.com/codecov/codecov-cli/blob/5d309ec1694c7853a06ed5f8b04a63f8705a5880/codecov_cli/helpers/ci_adapters/base.py#L101-L102) in [`github_actions.py`](https://github.com/codecov/codecov-cli/blob/5d309ec1694c7853a06ed5f8b04a63f8705a5880/codecov_cli/helpers/ci_adapters/github_actions.py).

### DIFF
--- a/codecov_cli/helpers/ci_adapters/github_actions.py
+++ b/codecov_cli/helpers/ci_adapters/github_actions.py
@@ -86,5 +86,8 @@ class GithubActionsCIAdapter(CIAdapterBase):
     def _get_service(self):
         return "github-actions"
 
+    def _get_git_service(self):
+        return "github"
+
     def get_service_name(self):
         return "GithubActions"


kdarkhan:
I just discovered https://github.com/codecov/feedback/issues/263

Based on https://github.com/codecov/feedback/issues/263#issuecomment-1929753219, my PR might fix the issue.

kdarkhan:
@drazisil-codecov could this PR fix the issue that you mentioned in [this comment](https://github.com/codecov/feedback/issues/263#issuecomment-1929753219)?

kdarkhan:
Thanks for response, @drazisil-codecov.

I will wait for confirmation from @thomasrockhu-codecov on the approach. If this approach is correct I can add tests as well. Did not want to invest too much work into this since I wasn't sure if this is the right fix.

PR #379, opened 2024-02-13
fix: better default file pattern for test results

joseph-sentry:


codecov_cli/services/upload/file_finder.py
```diff
@@ -36,7 +36,7 @@
 ]
 
 test_results_files_patterns = [
-    "*junit*",
+    "*junit.xml",
```
adrian-codecov, line 39: Ufffff, idk man, this looks like a heavy change, tread with caution

adrian-codecov (approved):

PR #395, opened 2024-03-04
Upload to staging as well in CI

joseph-sentry:


joseph-sentry:
> @joseph-sentry I think we should also have it run on prod as well as staging

I wasn't actually planning on making this an actual PR, I was just testing out the test results commenting on staging, I will clean up my commits then ask for a review!

PR #397, opened 2024-03-05
fix: better detect merge-commits

giovanni-guidini:
Apparently a common thing for CIs to do is create a merge-commit for changes
in a branch before running tests and stuff.
This means that - especially for not-directly-supported CIs - we would maybe
return a SHA for a commit that didn't exist in the branch.

These changes fix that by checking to see if the current commit is a merge commit.
If it is we return the second parent, the most recent commit before the current one.

Q: What if the current latest commit in the branch is a merge commit?
Well in this case the parent, which is also part of the branch, will have the coverage.

Users can still provide a commit sha value to override this behavior.

closes codecov/codecov-cli#372

### DIFF
--- a/codecov_cli/helpers/versioning_systems.py
+++ b/codecov_cli/helpers/versioning_systems.py
@@ -42,6 +42,20 @@ class GitVersioningSystem(VersioningSystemInterface):
 
     def get_fallback_value(self, fallback_field: FallbackFieldEnum):
         if fallback_field == FallbackFieldEnum.commit_sha:
+            # here we will get the commit SHA of the latest commit
+            # that is NOT a merge commit
+            p = subprocess.run(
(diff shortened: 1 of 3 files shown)


matt-codecov (approved):
do we know which CIs do this? how do we tend to handle testing behaviors we add for specific CIs? do we have demo accounts on all of them?

giovanni-guidini:
> do we know which CIs do this?

No, but I've been told "it's a common thing"

> how do we tend to handle testing behaviors we add for specific CIs? 

Specific CIs are typically the ones we directly support. We do unit tests, mostly.
The fallback values come from the CI's docs. The tests are [here](https://github.com/codecov/codecov-cli/tree/main/tests/ci_adapters)

> do we have demo accounts on all of them?

I highly doubt that.

Issue #405, opened 2024-04-04
Port over network filter and prefix


thomasrockhu-codecov:
The uploader currently supports [network prefix and filter](https://github.com/codecov/uploader/blob/371d90cded11c1de6df59f690e6c0df9d77cedc8/src/helpers/cli.ts#L108-L119) arguments. These should be added to the CLI for path fixing.

PR #406, opened 2024-04-05
feat: add network prefix and filter options

thomasrockhu-codecov:
fixes https://github.com/codecov/codecov-cli/issues/405

This adds the `--network-prefix` and `--network-filter` arguments that existed in the [uploader](https://github.com/codecov/uploader/blob/371d90cded11c1de6df59f690e6c0df9d77cedc8/src/helpers/cli.ts#L108-L119) (and the [bash uploader](https://github.com/codecov/codecov-bash/blob/master/codecov#L127-L129)).

This PR does the following:
- Alphabetizes some of the arguments
- Adds the `--network-prefix` and `--network-filter` arguments
- Pass those arguments down to `NetworkFinder.find_files` and execute on them
- Pass in the `--network-root-folder` to the `NetworkFinder.find_files` function also
- Change `_produce_file_fixes_for_network` to `_produce_file_fixes` as file fixes are not relevant to the network. This helps prevent file fixes from blowing up since the network filepaths are changing


giovanni-guidini (approved):

codecov_cli/commands/upload.py
```diff
@@ -181,29 +189,31 @@ def do_upload(
     ctx: click.Context,
     commit_sha: str,
     report_code: str,
+    branch: typing.Optional[str],
```
joseph-sentry, line 192: Curious to know, did you manually order the args alphabetically, or does your linter do it for you automatically, or did you have to use a separate tool?

codecov_cli/commands/upload.py
```diff
@@ -164,6 +164,14 @@ def _turn_env_vars_into_dict(ctx, params, value):
         default="coverage",
         type=click.Choice(["coverage", "test_results"]),
     ),
+    click.option(
+        "--network-filter",
+        help="Specify a filter on the files listed in the network section of the Codecov report. Useful for upload-specific path fixing",
```
joseph-sentry, line 169: I think it would be good to clarify here that the filter checks if the file path starts with the filtered string

codecov_cli/services/upload/upload_collector.py
```diff
@@ -169,7 +169,7 @@ def generate_upload_data(self, report_type="coverage") -> UploadCollectionResult
             network=network,
             files=files,
             file_fixes=(
-                self._produce_file_fixes_for_network(network)
+                self._produce_file_fixes([file.path for file in files])
```
joseph-sentry, line 172: I don't think this change is correct. `files` is the coverage files/ test result files and not the source code files, but I think we want to apply file fixes to the source code files which is the `network`

codecov_cli/commands/upload.py
thomasrockhu-codecov: manually

`:sort` in `vim`

codecov_cli/services/upload/upload_collector.py
thomasrockhu-codecov: I don't think this is correct. The `network` is a listing of files in the `git` directory (usually) and is used to map files in the report. File fixes, on the other hand, are specifically for files that are in the report, not in the network.

A reason this matters

```
| - a/
| --- __init__.py
| - b/
| --- __init__.py
```

In this project, there are 2 directories, both with `__init__.py` files. If I run tests only in folder `b/`, I will produce coverage results for just `b/__init__.py`. However, the report MAY only contain `__init__.py` as the filename, without the `b/`. 

The `network` will show
```
a/__init__.py
b/__init__.py
```
This will cause confusion for Codecov, because it will NOT know which `__init__.py` file to use.

As a result, the `network` section needs to be updated to only include files that start with `b/`.

thomasrockhu-codecov: From call, you are right, `files` is coverage files, not all files

codecov_cli/services/upload/upload_collector.py
```diff
@@ -150,9 +150,9 @@ def generate_upload_data(self, report_type="coverage") -> UploadCollectionResult
             prep.run_preparation(self)
         logger.debug("Collecting relevant files")
         network = self.network_finder.find_files()
```
joseph-sentry, line 152: Why are we not also ignoring files with the prefixes here?

thomasrockhu-codecov: @joseph-sentry because this list will be the input into the overall coverage report. We want the prefixes here because this is what will let us know, for example to only look at files in a particularly subdirectory of a monorepo.

joseph-sentry (approved):

PR #416, opened 2024-04-15
fix: add support for ssh url repos

### LINKED ISSUE #415: GitHub repo with ssh:// url does not work
Running codecovcli on a repo checked out from GitHub using a "ssh://git@github.com/organization/repo.git" is giving "405 Not Allowed" error when uploading the report.
Codecovcli was also logging:
`codecov": warning - 2024-03-26 13:45:52,145 -- Service not found: None. Possible services are ['github', 'gitlab', 'bitbucket', 'github_enterprise', 'gitlab_enterprise', 'bitbucket_server']`

escarls:
add ssh scheme to be parsed as service in parse_git_service()

Closes #415


joseph-sentry (changes requested):
Hi! Firstly, thank you for contributing to Codecov. 

This change looks good, I would just add the kind of URL we're now handling in this change to the the list of "possible cases" in the comment in the `parse_git_service` function.

escarls:
> This change looks good, I would just add the kind of URL we're now handling in this change to the the list of "possible cases" in the comment in the `parse_git_service` function.

I've updated the PR with the requested change.

joseph-sentry (approved):

joseph-sentry:
I approved. There is a policy on the repo that requires having all commits in a PR signed before we can merge, so after that's done we should be able to merge.

escarls:
Sorry for the delay @joseph-sentry and @thomasrockhu-codecov
I've never worked with signatures before, so I had to figure it out.
It was as you said, I had to rebase to add the signature.

escarls:
Anything else missing from my side before you can do the final review @joseph-sentry?

joseph-sentry (approved):

PR #428, opened 2024-04-29
feat: overwrite previous comment written by process-test-results

joseph-sentry:
Fixes: https://github.com/codecov/engineering-team/issues/1713


joseph-sentry:
> what is matrix doing here? Or rather, what's the point and what does it help do?

@thomasrockhu-codecov the idea is that the action would pass the `strategy.matrix` converted to JSON to the `--matrix` option so that the CLI knows what the matrix values currently are, so that there's only ever one comment for each matrix combination, otherwise we wouldn't be able to differentiate between comments made for different matrix runs in the same comment

joseph-sentry:
> @joseph-sentry is this specific to test-results? or will coverage need it too?

I think this only makes sense for test results right now, but in the future we could see if it could make sense for coverage as well. This might be a way to automate away flags

codecov_cli/commands/process_test_results.py
```diff
@@ -130,23 +145,38 @@ def process_test_results(
...
-        "Posting test results comment",
-    )
+    url = f"https://api.github.com/repos/{repo_slug}/issues/{pr_number}/comments"
+    # list comments
+    result = send_get_request(url=url, headers=headers)
+    comments = loads(result.text)
+    if matrix is not None:
+        matrix = matrix.strip()
+    for comment in comments:
+        if f"<!-- Codecov comment for {matrix} -->" in comment:
```
giovanni-guidini, line 165: Genius. Simple yet effective.

giovanni-guidini (commented):
So just for my understanding here.
Now instead of having 1 new comment per run (which is terribly spammy as we can see by this PR) we will query _all_ comments for the PR and find ours (more than one) and edit per matrix entry? So that if you have say 3 parallel test runners you get at most 3 comments?

How do you know what is the current entry in the matrix? (looks like you don't and just updates all comments, but is it possible to know?)

codecov_cli/commands/process_test_results.py
joseph-sentry: [This is not novel haha](https://github.com/actions-cool/maintain-one-comment). But i'm thinking we may want to mention the matrix combination this comment is for, in the comment.

joseph-sentry:
> Now instead of having 1 new comment per run (which is terribly spammy as we can see by this PR) we will query all comments for the PR and find ours (more than one) and edit per matrix entry? So that if you have say 3 parallel test runners you get at most 3 comments?

That is the goal

> How do you know what is the current entry in the matrix? (looks like you don't and just updates all comments, but is it possible to know?)

You are correct, that is a mistake on my part, I think the url will have to specify which comment i'm modifying

giovanni-guidini:
> That is the goal

Thanks for the clarification.

Can we be _more_ ambitious and maintain a single comment with maybe sections for each part of the matrix? Preferably expandable sections with the summary on top.

I see you apparently get the entire comment text, so it should be possible and you can use the comments to figure out the boundaries of the matrix pieces.

`comments = loads(result.text)`

I think that this would be preferable. Imagine you have 14 parallel runs. That's 14 comments, which is super spammy. But 1 is way easier.


For example, something like

-------
### Failed Test Results

<details><summary>:x: Python 3.12 • **`1 failed`**, 665 passed and 0 skipped.</summary>

| **Test Description** | **Failure message** |
| :-- | :-- |
| <pre>Testsuite:<br>api.temp.calculator.test_calculator::test_divide<br><br>Test name:<br>pytest<br></pre> | <pre>def<br>                test_divide():<br>                &amp;gt; assert Calculator.divide(1, 2) == 0.5<br>                E assert 1.0 == 0.5<br>                E + where 1.0 = &amp;lt;function Calculator.divide at 0x104c9eb90&amp;gt;(1, 2)<br>                E + where &amp;lt;function Calculator.divide at 0x104c9eb90&amp;gt; = Calculator.divide<br>                .../temp/calculator/test_calculator.py:30: AssertionError</pre> |
</details>

<details><summary>:white_check_mark: Python 3.11 • **`0 failed`**, 666 passed and 0 skipped.</summary></details>

<details><summary>:white_check_mark: Python 3.10 • **`0 failed`**, 666 passed and 0 skipped.</summary></details>

adrian-codecov (dismissed):
Looks good to me and read the comments. I'll let Gio weigh in as well as he's been checking this for a while too.

For what it's worth, the way BA handles knowing the comment is by doing a separate comment, tracking and persisting a different comment id. Dunno if helpful nor if this is the path to scale this, but just my 2 cents 👌

PR #430, opened 2024-05-05
fix: log info if a file shows up as a directory

thomasrockhu-codecov:
fixes https://github.com/codecov/codecov-action/issues/1303

`git ls-files` will pull submodules as files, this change just logs that a file shows up as such a directory and continues on

### DIFF
--- a/codecov_cli/services/upload/upload_collector.py
+++ b/codecov_cli/services/upload/upload_collector.py
@@ -140,9 +140,7 @@ class UploadCollector(object):
                 ),
             )
         except IsADirectoryError as err:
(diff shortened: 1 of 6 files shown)


codecov_cli/services/upload/upload_collector.py
```diff
@@ -139,6 +139,8 @@ def _get_file_fixes(
                     reason=err.reason,
                 ),
             )
+        except IsADirectoryError as err:
```
thomasrockhu-codecov, line 142: This is the relevant section

PR #434, opened 2024-05-06
feat: adjust for tokenless v3

joseph-sentry:
we are no longer sending the X-Tokenless-PR header and we are changing the format of the contents of the X-Tokenless header

The X-Tokenless header will now be set to `<fork user name>:<branch name>` and will be the only extra header we send for tokenless requests. We get the value of this header from the TOKENLESS environment variable that will be set by the Codecov action.

### DIFF
--- a/codecov_cli/commands/report.py
+++ b/codecov_cli/commands/report.py
@@ -13,6 +13,15 @@ logger = logging.getLogger("codecovcli")
 @click.option(
     "--code", help="The code of the report. If unsure, leave default", default="default"
 )
+@click.option(
+    "-P",
+    "--pr",
+    "--pull-request-number",
+    "pull_request_number",
+    help="Deprecated option, this option has no effect on the execution of this command anymore. Specify the pull request number mannually. Used to override pre-existing CI environment variables",
+    cls=CodecovOption,
+    fallback_field=FallbackFieldEnum.pull_request_number,
+)
 @global_options
 @click.pass_context
 def create_report(
@@ -23,6 +32,7 @@ def create_report(
     git_service: str,
     token: str,
     fail_on_error: bool,
+    pull_request_number: int,
 ):
     enterprise_url = ctx.obj.get("enterprise_url")
     logger.debug(
--- a/codecov_cli/commands/report.py
+++ b/codecov_cli/commands/report.py
@@ -13,15 +13,6 @@ logger = logging.getLogger("codecovcli")
 @click.option(
     "--code", help="The code of the report. If unsure, leave default", default="default"
 )
-@click.option(
-    "-P",
-    "--pr",
-    "--pull-request-number",
-    "pull_request_number",
-    help="Specify the pull request number mannually. Used to override pre-existing CI environment variables",
-    cls=CodecovOption,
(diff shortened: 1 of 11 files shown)


codecov_cli/commands/report.py
```diff
@@ -13,15 +13,6 @@
 @click.option(
     "--code", help="The code of the report. If unsure, leave default", default="default"
 )
-@click.option(
```
thomasrockhu-codecov, line 16: this is a breaking change. I think that users that run the CLI in a `workflow_dispatch` need this

joseph-sentry: I just don't know if it makes sense for the create-report command

thomasrockhu-codecov: ahhh I see, that does make sense. however, we should probably for now make it a `deprecated` field

codecov_cli/helpers/request.py
```diff
@@ -95,6 +96,15 @@ def get_token_header_or_fail(token: str) -> dict:
     return {"Authorization": f"token {token}"}
 
 
+def get_auth_header(token):
+    tokenless = os.environ.get("TOKENLESS", None)
```
thomasrockhu-codecov, line 100: are we making this change for the orb and step too? Why doesn't the CLI just do this itself?

joseph-sentry: i don't think the CLI has access to the same context as the action but i can double check, but you are correct we should make a similar change the the orb and step as well

codecov_cli/commands/report.py
```diff
@@ -18,7 +18,7 @@
     "--pr",
     "--pull-request-number",
     "pull_request_number",
-    help="Specify the pull request number mannually. Used to override pre-existing CI environment variables",
+    help="Deprecated option, this option has no effect on the execution of this command anymore. Specify the pull request number mannually. Used to override pre-existing CI environment variables",
```
thomasrockhu-codecov, line 21:
```suggestion
    help="[Deprecated], this option has no effect on the execution of this command anymore. Specify the pull request number mannually. Used to override pre-existing CI environment variables",
```

codecov_cli/commands/report.py
```diff
@@ -18,7 +18,7 @@
...
     cls=CodecovOption,
     fallback_field=FallbackFieldEnum.pull_request_number,
```
thomasrockhu-codecov, line 23: perhaps we don't need this anymore either

codecov_cli/services/commit/__init__.py
```diff
@@ -43,15 +42,7 @@ def create_commit_logic(
 def send_commit_data(
     commit_sha, parent_sha, pr, branch, slug, token, service, enterprise_url
 ):
-    decoded_slug = decode_slug(slug)
-    pull_dict = get_pull(service, decoded_slug, pr) if not token else None
-    if is_fork_pr(pull_dict):
-        headers = {"X-Tokenless": pull_dict["head"]["slug"], "X-Tokenless-PR": pr}
-        branch = pull_dict["head"]["slug"] + ":" + branch
```
matt-codecov, line 50: don't we still need this?

PR #458, opened 2024-06-13
support python 3.12

branchvincent:
This adds basic support for python 3.12:

- An unescaped `\` now [raises a SyntaxWarning](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes)
- Since 3.12 [removed setuptools](https://docs.python.org/3/whatsnew/3.12.html#ensurepip), `python setup.py` will fail to import setuptools. Note that directly executing setup.py has been [unofficially deprecated/discouraged](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/) for a while, so I've updated to use `pip` instead (there are other places as well but those might take more work)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,9 @@ jobs:
         with:
           submodules: true
           fetch-depth: 2
-      - uses: actions/setup-python@v3
+      - uses: actions/setup-python@v5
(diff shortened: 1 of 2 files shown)


giovanni-guidini (approved):
Thank you for this contribution 🙏

PR #461, opened 2024-06-21
fix: remove token check for report

thomasrockhu-codecov:


joseph-sentry (approved):

codecov_cli/services/report/__init__.py
```diff
@@ -80,7 +80,13 @@ def create_report_results_logic(
 def send_reports_result_request(
     commit_sha, report_code, encoded_slug, service, token, enterprise_url
 ):
-    headers = get_token_header_or_fail(token)
+    tokenless = os.environ.get("TOKENLESS")
```
joseph-sentry, line 83: I don't think the endpoint here supports this

codecov_cli/services/report/__init__.py
```diff
@@ -8,7 +8,7 @@
 from codecov_cli.helpers.config import CODECOV_API_URL
 from codecov_cli.helpers.encoder import decode_slug, encode_slug
 from codecov_cli.helpers.request import (
-    get_token_header_or_fail,
+    get_token_header,
```
joseph-sentry, line 11: we should still keep get_token_header_or_fail here

joseph-sentry (changes requested):

joseph-sentry (approved):

codecov_cli/services/report/__init__.py
giovanni-guidini: Your comment is conflicting with the code

giovanni-guidini (approved):

codecov_cli/services/report/__init__.py
joseph-sentry: there was an old version of the PR where `get_token_header_or_fail` was being deleted, i meant that we should keep both

PR #464, opened 2024-06-25
🐛 Use config-set token as CLI default fallback

webknjaz:
Just like in the old days... Before this patch, the v4-related rewrite of the Codecov uploader CLI lost the ability to use the token set in the config file [[1]]. This change fixes the regression recovering the original behavior.

The bug has been discovered during the re-configuration of ``codecov/codecov-action`` in ``pytest-dev/pytest``. We use the clear-text token to improve stability of uploads in PRs from forks so it makes sense to keep it in the config rather than in the GitHub Actions CI/CD definition file.

[1]: https://docs.codecov.com/docs/codecovyml-reference#codecovtoken

### DIFF
--- a/codecov_cli/main.py
+++ b/codecov_cli/main.py
@@ -58,6 +58,8 @@ def cli(
     ctx.obj["codecov_yaml"] = load_cli_config(codecov_yml_path)
     if ctx.obj["codecov_yaml"] is None:
         logger.debug("No codecov_yaml found")
+    elif (token := ctx.obj["codecov_yaml"].get("codecov", {}).get("token")) is not None:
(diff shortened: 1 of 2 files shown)


webknjaz:
Hey @thomasrockhu-codecov, I've found another v4 regression FYI ^

joseph-sentry (approved):

webknjaz:
I was confused by these failures on `main`: https://github.com/codecov/codecov-cli/actions/runs/9686076902/job/26727674869#step:6:14. But then, I saw they existed before merging this PR as well.

webknjaz:
@joseph-sentry @thomasrockhu-codecov would you mind making a new release with this patch included?

PR #465, opened 2024-06-25
🐛 Ensure that no-token case is actually logged

webknjaz:
Prior to this patch, there was an unreachable code path that seems to have been intended for marking the case when no token is provided with the `"NOTOKEN"` string in the debug log output. However the logic behind choosing to show it contained two opposite checks. Said place in the logging utils, didn't have any coverage either.

This change updates the code to make all of its branches reachable. Moreover, it adds test cases for all the branches that remain in the implementation.


webknjaz:
FTR I noticed this unreachable path while working on #464.

webknjaz:
:crossed_fingers:

webknjaz:
😮‍💨 I'll look into it later..

PR #473, opened 2024-07-15
chore: consistent naming of upload options

giovanni-guidini:
Makes the `--report-code` of `do-upload` to also accept
`--code`. This makes it consistent with other commands.

closes codecov/feedback#380


Swatinem (dismissed):

Swatinem (approved):

.github/workflows/build_assets.yml
```diff
@@ -31,8 +31,7 @@ jobs:
           - os: ubuntu-20.04
             TARGET: ubuntu
             CMD_REQS: >
-              pip install -r requirements.txt
-              pip install .
+              pip install -r requirements.txt && pip install .
```
webknjaz, line 34: @giovanni-guidini FYI, both deps could be passed to the same `pip install` command. There's no technical reason for running it twice here.

PR #475, opened 2024-07-15
fix: accept tokenless value from branch instead of env var

joseph-sentry:
depends on: https://github.com/codecov/codecov-action/pull/1511
fixes: https://github.com/codecov/engineering-team/issues/2065

### DIFF
--- a/codecov_cli/services/commit/__init__.py
+++ b/codecov_cli/services/commit/__init__.py
@@ -53,15 +53,21 @@ def send_commit_data(
     enterprise_url,
     args,
 ):
-    # this is how the CLI receives the username of the user to whom the fork belongs
-    # to and the branch name from the action
-    tokenless = os.environ.get("TOKENLESS")
-    if tokenless:
-        headers = None  # type: ignore
-        branch = tokenless  # type: ignore
-        logger.info("The PR is happening in a forked repo. Using tokenless upload.")
+    # Old versions of the GHA use this env var instead of the regular branch
+    # argument to provide an unprotected branch name
+    if tokenless := os.environ.get("TOKENLESS"):
+        branch = tokenless
+
+    if branch and ":" in branch:
+        logger.info(f"Creating a commit for an unprotected branch: {branch}")
+    elif token is None:
+        logger.warning(
+            f"Branch `{branch}` is protected but no token was provided\nFor information on Codecov upload tokens, see https://docs.codecov.com/docs/codecov-tokens"
+        )
     else:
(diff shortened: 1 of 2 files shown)


giovanni-guidini (dismissed):

codecov_cli/services/commit/__init__.py
```diff
@@ -45,10 +45,8 @@ def send_commit_data(
...
     # this is how the CLI receives the username of the user to whom the fork belongs
     # to and the branch name from the action
-    tokenless = os.environ.get("TOKENLESS")
-    if tokenless:
+    if not token and branch and ":" in branch:
         headers = None  # type: ignore
-        branch = tokenless  # type: ignore
         logger.info("The PR is happening in a forked repo. Using tokenless upload.")
     else:
         headers = get_token_header_or_fail(token)
```
matt-codecov, line 52: strictly speaking, we no longer need any of this logic anymore. we just send whatever we have for the token and the branch, and the API will return a 401 if it's not allowed

the logs may still be helpful, so maybe change to:
```
if not token:
    if branch and ":" in branch:
        logger.info("Creating a commit on an unprotected branch")
    else:
        logger.warning("Token is missing, but branch is missing or protected")

headers = get_token_header(token)
```

matt-codecov (changes requested):
sorry for the whiplash :(

matt-codecov (approved):

tony-codecov (approved):

codecov_cli/services/commit/__init__.py
```diff
@@ -53,15 +53,21 @@ def send_commit_data(
...
-    # to and the branch name from the action
-    tokenless = os.environ.get("TOKENLESS")
-    if tokenless:
-        headers = None  # type: ignore
-        branch = tokenless  # type: ignore
-        logger.info("The PR is happening in a forked repo. Using tokenless upload.")
+    # Old versions of the GHA use this env var instead of the regular branch
+    # argument to provide an unprotected branch name
+    if tokenless := os.environ.get("TOKENLESS"):
+        branch = tokenless
```
thomasrockhu-codecov, line 59: branch = os.environ.get("TOKENLESS", None) could have worked too

matt-codecov: that would clobber whatever was already in branch

thomasrockhu-codecov: @matt-codecov ahhh that's a good point

Issue #469, opened 2024-07-01
Is the verbose mode (`-v`) supposed to let the clear-text token into the output of `codecov -v create-commit`?


webknjaz:
Here https://github.com/pytest-dev/pytest/actions/runs/9746387730/job/26896668869?pr=12553#step:9:66 I noticed that one debug log entry has it sanitized, and the other reveals the token value within an escaped JSON string. I haven't attempted debugging why this is happening, but it seems it might be related to said escaping.
It's not very problematic in this specific project, though.

P.S. Your own dogfooding step @ https://github.com/codecov/codecov-cli/actions/runs/9715039688/job/26815507245#step:6:2 does not manifest this problem since it only invokes `do-upload` and doesn't run `create-commit` as the action wrapper does.
P.P.S. Many people use `${{ secrets.CODECOV_TOKEN }}` in the GHA realm, so GHA masks the value for them, on a different level. But when the value is not coming from a secret, it's not hidden by the platform.

Issue #468, opened 2024-07-01
Mark the token for masking in GHA logs


webknjaz:
This is almost the same as https://github.com/codecov/codecov-action/issues/1498, except the action only knows of things incoming through the action inputs interface. But `codecov-cli` has a better context available:
1) It can be called @ GHA directly w/o that action and should still be able to detect the environment where it's being executed
2) It can read token from a config file and `codecov/codecov-action` wouldn't know about that so it wouldn't be able to mask it

Action item:
* [ ] `print(f'::add-mask::{token}', file=sys.stderr)` per https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-masking-a-string as soon as it's known in the program

PR #485, opened 2024-08-15
fix: sanitize yaml token from logs

thomasrockhu-codecov:
fixes https://github.com/codecov/codecov-cli/issues/469
fixes https://github.com/codecov/codecov-cli/issues/468
fixes https://github.com/codecov/codecov-action/issues/1498

### DIFF
--- a/codecov_cli/helpers/request.py
+++ b/codecov_cli/helpers/request.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from sys import exit
 from time import sleep
@@ -143,7 +144,9 @@ def log_warnings_and_errors_if_any(
     )
     logger.debug(
         f"{process_desc} result",
-        extra=dict(extra_log_attributes=dict(result=sending_result)),
+        extra=dict(
+            extra_log_attributes=dict(result=_sanitize_request_result(sending_result))
+        ),
     )
     if sending_result.warnings:
         number_warnings = len(sending_result.warnings)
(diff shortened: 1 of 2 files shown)


codecov_cli/helpers/request.py
```diff
@@ -157,3 +160,27 @@ def log_warnings_and_errors_if_any(
...
+
+def _sanitize_request_result(result: RequestResult):
+    if not hasattr(result, "text"):
+        return result
+
+    try:
+        text_as_dict = json.loads(result.text)
+        token = text_as_dict.get("repository").get("yaml").get("codecov").get("token")
+        if token:
+            sanitized_token = str(token)[:1] + 18 * "*"
```
matt-codecov, line 173: why include the first character?

matt-codecov (dismissed):

codecov_cli/helpers/request.py
thomasrockhu-codecov: @matt-codecov we do this in other parts of the system. the intention is that it helps to diagnose issues (e.g. copying `CODECOV_TOKEN` as the token)

matt-codecov (approved):

webknjaz:
@thomasrockhu-codecov could you document why you decided not to mark the token as a secret using GitHub Actions workflow commands? That would prevent it from being printed out if leaked through other commands that follow the codecov-cli invocation.

PR #496, opened 2024-09-05
chore: typehint context object

giovanni-guidini:


Swatinem (dismissed):
lgtm, though it looks like some older python versions don’t like this?

Swatinem (approved):

PR #506, opened 2024-09-20
feat: collect additional params from cli usage

thomasrockhu-codecov:
fixes https://github.com/codecov/infrastructure-team/issues/441


joseph-sentry (approved):
seems ok to me, it feels bad that we have to edit each file 1 by 1 to add such a thing but i guess it makes sense since each one passes different data to its post request

i still think we should at least remove `folders_to_ignore` since we wouldn't ever have access to that information anyways, but i trust your judgement

PR #507, opened 2024-09-24
Aggregate the test results files in CI

michelletran-codecov:
Update the CI to aggregate results all into one comment. This will be useful for extracting into its own action in the future.

Basically, changes needed to ensure one comment:
- test junit.xml files are exported to an artifact of some `${UNIQUE_IDENTIFIER}junit.xml`
- an "aggregation" job that puts all the `*junit.xml` files into a directly
- calling process_test_results with the dir with all the junit files

I've added a breaking test for testing purposes. Once I get an approval, I'll remove it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,12 +76,18 @@ jobs:
         pip install -r tests/requirements.txt
     - name: Test with pytest
       run: |
-        pytest --cov --junitxml=junit.xml
+        pytest --cov --junitxml=${{matrix.python-version}}junit.xml
     - name: Dogfooding codecov-cli
       if: ${{ !github.event.pull_request.head.repo.fork && github.repository_owner == 'codecov' }}
       run: |
         codecovcli -v do-upload --fail-on-error -t ${{ secrets.CODECOV_TOKEN }} --plugin pycoverage --flag python${{matrix.python-version}}
         codecovcli do-upload --report-type test_results --fail-on-error -t ${{ secrets.CODECOV_TOKEN }} --plugin pycoverage --flag python${{matrix.python-version}}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -76,12 +76,18 @@ jobs:
...
       run: |
-        pytest --cov --junitxml=junit.xml
+        pytest --cov --junitxml=${{matrix.python-version}}junit.xml
     - name: Dogfooding codecov-cli
       if: ${{ !github.event.pull_request.head.repo.fork && github.repository_owner == 'codecov' }}
       run: |
         codecovcli -v do-upload --fail-on-error -t ${{ secrets.CODECOV_TOKEN }} --plugin pycoverage --flag python${{matrix.python-version}}
         codecovcli do-upload --report-type test_results --fail-on-error -t ${{ secrets.CODECOV_TOKEN }} --plugin pycoverage --flag python${{matrix.python-version}}
+    - name: Upload artifacts for test-results-processing
+      if: ${{ always() }}
```
Swatinem, line 86: do you need this because the previous `run` command is optional?

Swatinem (approved):

.github/workflows/ci.yml
michelletran-codecov: No. The reason for it is because the `Test with pytest` was failing the build and all subsequent steps were cancelled. Since the test processing is expecting failures, we need this step to run regardless of test failures.

Swatinem (approved):

.github/workflows/ci.yml
joseph-sentry: I think using `!cancelled()` may be better here because I think `always()` will make it so the step will run even if the job is cancelled

PR #533, opened 2024-10-15
Remove token enforcement for true tokenless endpoints

nora-codecov:
Went through the upload endpoints on API:

- endpoints where we want to allow tokenless: I've made sure the method uses `get_token_header()` which allows `None` for the `token`, which will allow tokenless requests through to the api.

- endpoints where we don't want to allow tokenless: I've made sure they use `get_token_header_or_fail()` which does not allow `None`, so the request will be kicked out before it gets to api. This is not required for the tokenless logic to work, but is nice to have since it prevents junk requests from clogging api.

- update: we are going to allow tokenless uploads on all upload related endpoints, so switched them all to `get_token_header()`.  `get_token_header_or_fail()` is now unused, but I'm leaving it in code for if we need to switch an endpoint back to token required.

https://github.com/codecov/engineering-team/issues/2300


setup.py
```diff
@@ -26,6 +26,7 @@
         "tree-sitter==0.20.*",
         "test-results-parser==0.5.*",
         "regex",
+        "pytest-mock"
```
thomasrockhu-codecov, line 29: does this need to be an install requirement?

codecov_cli/services/commit/__init__.py
```diff
@@ -19,7 +19,7 @@ def create_commit_logic(
     pr: typing.Optional[str],
     branch: typing.Optional[str],
     slug: typing.Optional[str],
-    token: str,
+    token: typing.Union[str, None],
```
michelletran-codecov, line 22: How about using Optional type instead? (and all the other places with Union)
```suggestion
    token: typing.Optional[str],
```

codecov_cli/services/report/__init__.py
```diff
@@ -118,7 +116,7 @@ def send_reports_result_get_request(
     enterprise_url,
     fail_on_error=False,
 ):
-    headers = get_token_header_or_fail(token)
+    headers = get_token_header(token)
```
michelletran-codecov, line 119: `get_token_header` currently doesn't take an optional argument ([according to the type annotation](https://github.com/codecov/codecov-cli/blob/6efe50c4d4c4b51a9c7799e48c1042288ba64378/codecov_cli/helpers/request.py#L113)). Care to fix that as well? :p

michelletran-codecov (commented):
Generally LGTM. Just a few comments about the type annotation.

codecov_cli/services/commit/__init__.py
nora-codecov: sure!

codecov_cli/services/report/__init__.py
nora-codecov: certainly haha

thomasrockhu-codecov (dismissed):
just fix the lint

thomasrockhu-codecov (approved):

PR #545, opened 2024-10-30
fix: pass args to send_reports_result_request

joseph-sentry:


codecov_cli/services/report/__init__.py
```diff
@@ -3,6 +3,8 @@
 import time
 import typing
 
+import requests
```
thomasrockhu-codecov, line 6: not needed?

thomasrockhu-codecov (approved):

PR #551, opened 2024-11-06
Upload-process uses upload-coverage endpoint

tony-codecov:
This PR adds a `upload-coverage` command to use a new combined upload approach that consolidates the create commit, create report, and upload steps into a single network call when possible. This optimization reduces network overhead by combining what were previously three separate API calls into one.

- Added new `upload_coverage` function that handles all three steps (create commit, create report, and upload) in one operation. This will be called with the `upload-coverage` command.

`upload-process` still behaves the same as previously.

Depends on https://github.com/codecov/codecov-api/pull/962

Closes https://github.com/codecov/engineering-team/issues/2538


codecov_cli/services/upload/upload_sender.py
```diff
@@ -176,9 +185,15 @@ def get_url_and_possibly_update_data(
...
+        combined_upload=False,
     ):
         if report_type == "coverage":
-            url = f"{upload_url}/upload/{git_service}/{encoded_slug}/commits/{commit_sha}/reports/{report_code}/uploads"
+            if combined_upload:
+                url = (
+                    f"{upload_url}/upload/{git_service}/{encoded_slug}/combined-upload"
+                )
+            else:
+                url = f"{upload_url}/upload/{git_service}/{encoded_slug}/commits/{commit_sha}/reports/{report_code}/uploads"
```
Swatinem, line 196: the formatting looks a bit weird. how about this:
```suggestion
            root_url = f"{upload_url}/upload/{git_service}/{encoded_slug}/"
            if combined_upload:
                url = f"{root_url}/combined-upload"
            else:
                url = f"{root_url}/commits/{commit_sha}/reports/{report_code}/uploads"
```

Swatinem (dismissed):

codecov_cli/services/upload/upload_sender.py
tony-codecov: This looks great, thanks for the suggestion!

Swatinem (dismissed):

Swatinem (approved):

Swatinem (approved):

PR #568, opened 2024-11-28
fix: make default search case insensitive

joseph-sentry:
when looking for files in the default patterns search, it's better to
be flexible and not care about case sensitive matching. For example,
certain frameworks output their test results with the word "test" in
all upper case and we shouldn't need to specify both "test" and "TEST"
to match those files.

### DIFF
--- a/codecov_cli/helpers/folder_searcher.py
+++ b/codecov_cli/helpers/folder_searcher.py
@@ -85,7 +85,9 @@ def search_files(
                     yield file_path
 
 
-def globs_to_regex(patterns: List[str]) -> Optional[Pattern]:
+def globs_to_regex(
+    patterns: List[str], case_sensitive: bool = True
+) -> Optional[Pattern]:
     """
     Converts a list of glob patterns to a combined ORed regex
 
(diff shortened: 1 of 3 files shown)


codecov_cli/helpers/folder_searcher.py
```diff
@@ -100,4 +102,8 @@ def globs_to_regex(patterns: List[str]) -> Optional[Pattern]:
         return None
 
     regex_str = ["(" + translate(pattern) + ")" for pattern in patterns]
-    return re.compile("|".join(regex_str))
+    # return re.compile("|".join(regex_str))
```
Swatinem, line 105:
```suggestion
```

Swatinem (dismissed):

Swatinem (approved):

thomasrockhu:
I don't think this should go in, it will change user behavior for search and I expect it will have unexpected consequences. I'd rather users have to be explicit about the files they want then not be able to not include it

joseph-sentry:
good point, i will close

Issue #570, opened 2024-12-05
Update CLI package for pypi


thomasrockhu-codecov:
This email is notifying you of an upcoming deprecation that we have determined may affect you as a result of your recent upload to 'codecov-cli'.

In the future, PyPI will require all newly uploaded source distribution filenames to comply with [PEP 625](https://peps.python.org/pep-0625/). Any source distributions already uploaded will remain in place as-is and do not need to be updated.

Specifically, your recent upload of 'codecov-cli-9.1.0.tar.gz' is incompatible with PEP 625 because it does not contain the normalized project name 'codecov_cli'.

In most cases, this can be resolved by upgrading the version of your build tooling to a later version that supports PEP 625 and produces compliant filenames.

PR #577, opened 2024-12-09
fix: change name to codecov_cli for PEP625

thomasrockhu-codecov:
fixes https://github.com/codecov/codecov-cli/issues/570

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ with open(path.join(here, "README.md"), encoding="utf-8") as f:
     long_description = f.read()
 
 setup(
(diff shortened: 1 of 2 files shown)


trent-codecov (dismissed):

setup.py
```diff
@@ -9,7 +9,7 @@
     long_description = f.read()
 
 setup(
-    name="codecov-cli",
+    name="codecov_cli",
```
drazisil-codecov, line 12: Does this change the binary name, or only the package name?

trent-codecov (approved):

Issue #641, opened 2025-02-19
TypeError: object of type 'NoneType' has no len() when uploading coverage


danedmunds:
Hi there! 

We've noticed that since release 10.1.0 went out that our codecov upload github action is failing with the following stack trace

```
 Traceback (most recent call last):
  File "codecov_cli/main.py", line 97, in <module>
  File "codecov_cli/main.py", line 93, in run
  File "click/core.py", line 1161, in __call__
  File "click/core.py", line 1082, in main
  File "click/core.py", line 1697, in invoke
  File "click/core.py", line 1443, in invoke
  File "click/core.py", line 788, in invoke
  File "click/decorators.py", line 33, in new_func
  File "codecov_cli/commands/upload_coverage.py", line 85, in upload_coverage
  File "click/core.py", line 788, in invoke
  File "codecov_cli/services/upload_coverage/__init__.py", line 51, in upload_coverage_logic
  File "codecov_cli/services/upload/__init__.py", line 102, in do_upload_logic
  File "codecov_cli/services/upload/upload_collector.py", line 169, in generate_upload_data
TypeError: object of type 'NoneType' has no len()
```

I suspect this is due to this change https://github.com/codecov/codecov-cli/pull/625
I'm not very well versed in python but maybe len(None) isn't allowed?

danedmunds:
Thanks @joseph-sentry ! I didn't have the python chops to help out :)

webknjaz:
@thomasrockhu-codecov @joseph-sentry FYI, you should be able to mark the PyPI release as yanked to limit the damage.

webknjaz:
> [@thomasrockhu-codecov](https://github.com/thomasrockhu-codecov) [@joseph-sentry](https://github.com/joseph-sentry) FYI, you should be able to mark the PyPI release as yanked to limit the damage.

UPD: I see it got yanked which is why there's no mass reports with this traceback.

webknjaz:
Oh, I'm still seeing this in CIs because the actions download 10.1.0 from another hosting: https://github.com/ansible/awx-plugins/actions/runs/13437325559/job/37542819968#step:24:41.

Interestingly, there's some API that returns HTML in place of JSON: https://github.com/ansible/awx-plugins/actions/runs/13437325559/job/37542819968#step:23:45. But it's probably being hit on the action side, not the CLI project...

@thomasrockhu-codecov any chance to change “latest” to the previous version there too?

[...]

PR #643, opened 2025-02-19
fix: len(None) bug

joseph-sentry:
when the versioning system is `None`, `list_relevant_files` returns `None` so this changes the code in the upload_collector and the network_finder to handle that case

Fixes: https://github.com/codecov/codecov-cli/issues/641

### DIFF
--- a/codecov_cli/services/upload/network_finder.py
+++ b/codecov_cli/services/upload/network_finder.py
@@ -20,7 +20,7 @@ class NetworkFinder(object):
     def find_files(self, ignore_filters=False) -> typing.List[str]:
         files = self.versioning_system.list_relevant_files(self.network_root_folder)
 
-        if not ignore_filters:
+        if files and not ignore_filters:
             if self.network_filter:
                 files = [file for file in files if file.startswith(self.network_filter)]
             if self.network_prefix:
--- a/codecov_cli/services/upload/upload_collector.py
+++ b/codecov_cli/services/upload/upload_collector.py
@@ -166,7 +166,7 @@ class UploadCollector(object):
                 f"Found {len(report_files)} {report_type.value} files to report"
             )
             logger.debug(
-                f"Found {len(network)} network files to report, ({len(unfiltered_network)} without filtering)"
+                f"Found {len(network) if network else None} network files to report, ({len(unfiltered_network) if network else None} without filtering)"
             )
             if not report_files:
                 if report_type == ReportType.TEST_RESULTS:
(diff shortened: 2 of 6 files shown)


nora-codecov (approved):

codecov_cli/services/upload/network_finder.py
```diff
@@ -20,7 +20,7 @@ def __init__(
     def find_files(self, ignore_filters=False) -> typing.List[str]:
```
Dreamsorcerer, line 20: Is mypy run on this project?
If files is None (as described in the PR description), your changed code surely results in returning None, in which case this annotation is wrong and there's a good chance another part of the codebase could mess up when expecting a list.

Dreamsorcerer: OK, I see the root problem (which confirms to me that this is probably not being checked by mypy, as it should have errored here).

The method is defined on the base class, but it's not an abstract method. So, if inherited (on instantiated directly), this returns None, even though the annotation says it returns a list (which is a clear type error that should be caught by any type checker):
https://github.com/codecov/codecov-cli/blob/63bd5f2697c6adbaecef0d6e442c952b2fb7cb87/codecov_cli/helpers/versioning_systems.py#L23-L24

I think the correct fix is to either replace that `pass` with a `return []`, so it actually does the correct behaviour, or to turn those methods into abstract methods and require the subclass to define them correctly.

joseph-sentry: Thanks for the feedback, I've implemented the ABC / abstractmethod suggestion.

codecov_cli/services/upload/upload_collector.py
```diff
@@ -166,7 +166,7 @@ def generate_upload_data(
                 f"Found {len(report_files)} {report_type.value} files to report"
             )
             logger.debug(
-                f"Found {len(network)} network files to report, ({len(unfiltered_network)} without filtering)"
+                f"Found {len(network) if network else None} network files to report, ({len(unfiltered_network) if network else None} without filtering)"
```
thomasrockhu-codecov, line 169:
```suggestion
                f"Found {len(network) if network else 0} network files to report, ({len(unfiltered_network) if network else 0} without filtering)"
```

tests/services/upload/test_upload_collector.py
```diff
@@ -1,7 +1,10 @@
 from pathlib import Path
 from unittest.mock import patch
 
-from codecov_cli.helpers.versioning_systems import GitVersioningSystem
+from codecov_cli.helpers.versioning_systems import (
+    GitVersioningSystem,
+    NoVersioningSystem,
```
thomasrockhu-codecov, line 6: this is imported and not used

thomasrockhu-codecov: ignore, I can't read

thomasrockhu-codecov (approved):

nora-codecov (approved):

PR #644, opened 2025-02-20
Remove ATS-related configuration

Swatinem:


Swatinem:
> We will also want to remove the associated tests

Eventually yes. All the code within the CLI, and associated tests.

I’m not sure if we have any kind of tests depending on the actual upload of ATS data for the CLI?

Fully deprecating this feature will involve announcing the removal, and hiding the related commands from the CLI for some time before actually removing them completely.

PR #650, opened 2025-02-25
add --toc-recurse-submodules argument

matt-codecov:
`git ls-files` does not recurse into submodules unless you pass `--recurse-submodules`. we will need to do that while migrating to a monorepo, so i added a CLI flag to control whether that argument is passed to `git ls-files`

i think this will also need a corresponding change in the wrapper and action


codecov_cli/commands/upload.py
```diff
@@ -272,6 +279,7 @@ def do_upload(
                 build_code=build_code,
                 build_url=build_url,
                 commit_sha=commit_sha,
+                toc_recurse_submodules=toc_recurse_submodules,
```
thomasrockhu-codecov, line 282: this list is alphabetized for sanity

README.md
```diff
@@ -170,6 +170,7 @@ Codecov-cli supports user input. These inputs, along with their descriptions and
 |-s, --dir, --coverage-files-search-root-folder | Folder where to search for coverage files default: (Current Working Directory) | Optional
 |--exclude, --coverage-files-search-exclude-folder | Folders to exclude from search | Optional
 |-f, --file, --coverage-files-search-direct-file | Explicit files to upload | Optional
+|--toc-recurse-submodules | Whether to enumerate files inside of submodules for path-fixing purposes. Off by default. | Optional
```
thomasrockhu-codecov, line 173: how do you feel about just `--recurse-submodules`

thomasrockhu-codecov (changes requested):

codecov_cli/commands/upload.py
matt-codecov: will update

README.md
matt-codecov: i wanted the name to be more indicative of its purpose, but i guess `toc` is only meaningful to us. will change

thomasrockhu-codecov (approved):

PR #654, opened 2025-03-10
Add create commit to `empty-upload` command

michelletran-codecov:
This is so that we can also create the commit before uploading. The create-commit is optional (i.e. if users don't pass in values, then their workflow shouldn't break). It will only create a commit if the values are passed.


codecov_cli/commands/empty_upload.py
```diff
@@ -26,11 +46,30 @@ def empty_upload(
...
+    parent_sha: typing.Optional[str],
+    pull_request_number: typing.Optional[int],
+    branch: typing.Optional[str],
 ):
     with sentry_sdk.start_transaction(op="task", name="Empty Upload"):
         with sentry_sdk.start_span(name="empty_upload"):
             enterprise_url = ctx.obj.get("enterprise_url")
             args = get_cli_args(ctx)
+
+            if parent_sha and pull_request_number and branch:
```
thomasrockhu-codecov, line 58: is this necessary to check for? I would assume that not every empty_upload would be on a PR for example

michelletran-codecov: Oh true. I added these here to check if I should create a commit. I don't want to call create commit _if_ the user is not intending to create a commit (i.e. old workflow). But these fields were all optional for the `create-commit` command anyways, so maybe it's fine to remove the check. I'll check on this.

michelletran-codecov: I think we can remove the check. The create-commit logic seems to check that a commit exists before it adds, so it doesn't create duplicate updates.

thomasrockhu-codecov (approved):

PR #663, opened 2025-03-17
Make `label-analysis` and `static-analysis` noops

Swatinem:
This rips out all of `static-analysis` including building treesitter languages and processing source files.

Ripping out `label-analysis` works rather by running the fallback code unconditionally without making requests to the backend API.

---

I would still wait with this a little until we have the final go-ahead in https://github.com/codecov/engineering-team/issues/3364


codecov_cli/commands/staticanalysis.py
```diff
@@ -60,27 +56,4 @@ def static_analysis(
     force,
     folders_to_exclude: typing.List[pathlib.Path],
 ):
-    with sentry_sdk.start_transaction(op="task", name="Static Analysis"):
```
thomasrockhu-codecov, line 63: we probably still want to capture this for usage

requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.12
```
thomasrockhu-codecov, line 2: I think we want to keep this as 3.11, but not 100% positive

thomasrockhu-codecov (changes requested):
I'd like to split the build changes from the noop changes in a different PR. We've had issues with our build system already recently and I'd like to ensure that these changes will still produce the correct assets.

Swatinem:
I split out the build system related changes to https://github.com/codecov/codecov-cli/pull/664

thomasrockhu-codecov (dismissed):

thomasrockhu-codecov (approved):

PR #667, opened 2025-03-25
Test on all OSs

Swatinem:
This adds Windows and macOS to our CI test matrix.

It also changes quite a lot of internal functionality and tests to use `Path.as_posix()` which always prints paths with forward slashes.
This was primarily done to make a ton of tests work on Windows, but IMO it is also a good idea to normalize paths internally to use forward slashes.

---

As https://github.com/codecov/codecov-cli/pull/661 is primarily an issue on Windows, it would be nice to run tests on Windows as well.
And also macOS as well, why not. This should cover the major OSs that folks will be running in their own CI jobs that are using the CLI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,16 +70,12 @@ jobs:
           codecovcli create-report -t ${{ secrets.CODECOV_TOKEN }} --git-service github
 
   build-test-upload:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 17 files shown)


joseph-sentry (approved):
i think there's some required checks settings in the repo which is causing the CI to expect certain checks with diff names @thomasrockhu-codecov do you have perms to change those?

PR #683, opened 2025-05-01
fix: handle merge_group github events

joseph-sentry:
if we get a branch with gh-readonly-queue, it comes from a merge queue and the actual branch name is located in the middle


codecov_cli/helpers/ci_adapters/github_actions.py
```diff
@@ -67,20 +67,22 @@ def _get_slug(self):
...
+            match = re.search(r"refs/heads/(.*)", branch)
+            if match is None:
+                return None
+            branch = match.group(1)
 
-        if match is None:
-            return None
+            # handling for merge queue CI runs
+            if branch and branch.startswith("gh-readonly-queue/"):
+                return branch.split("/")[1]
```
thomasrockhu-codecov, line 84: for clarity can we do

```suggestion
            if branch:
                branch = branch.split("/")[1] if branch.startswith("gh-readonly-queue/") else branch
                return branch
```

thomasrockhu-codecov (dismissed):
logic is sound

codecov_cli/helpers/ci_adapters/github_actions.py
```diff
@@ -66,21 +66,21 @@ def _get_slug(self):
...
-            return None
-
-        match = re.search(r"refs/heads/(.*)", branch_ref)
+        branch = head_ref or ref
 
-        if match is None:
-            return None
+        # branch format for merge queue CI runs:
+        # gh-readonly-queue/<branch-name>/<pr-number>-<pr-name>
+        if branch.startswith("gh-readonly-queue/"):
```
thomasrockhu-codecov, line 81: I think this will be angry if it's still `None`

thomasrockhu-codecov (changes requested):

codecov_cli/helpers/ci_adapters/github_actions.py
```diff
@@ -66,21 +66,21 @@ def _get_slug(self):
...
+            return branch.split("/")[1]
 
-        branch = match.group(1)
         return branch or None
```
thomasrockhu-codecov, line 84: my worry here is if we depend on `None` to be meaningful versus `''`. Previous logic doesn't account for empty string

thomasrockhu-codecov (dismissed):

thomasrockhu-codecov (commented):
@joseph-sentry this looks good, can we move it to `prevent-cli` though?

PR #702, opened 2026-04-21
feat: Send upload size to upload request and cli version with API requests

calvin-codecov:
Closes https://github.com/codecov/core-engineering/issues/31

- Changes upload size calculation from sys.getsizeof() to len() as len more accurately represents the bytes' size while getsizeof includes Python overhead in its memory size.
- Include Codecov CLI version in the data for all post requests
- Send through correct file_not_found data attribute

### DIFF
--- a/codecov-cli/codecov_cli/services/upload/upload_sender.py
+++ b/codecov-cli/codecov_cli/services/upload/upload_sender.py
@@ -2,7 +2,6 @@ import base64
 import json
 import logging
 import typing
-import sys
 import zlib
 from typing import Any, Dict
 
@@ -97,6 +96,8 @@ class UploadSender(object):
                 reports_payload = self._generate_payload(
                     upload_data, env_vars, report_type
                 )
+                reports_payload_size = len(reports_payload)
+                data["report_payload_bytes"] = reports_payload_size
 
             with sentry_sdk.start_span(name="upload_sender_storage_request"):
                 logger.debug("Sending upload request to Codecov")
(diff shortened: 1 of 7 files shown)


codecov-cli/tests/helpers/test_upload_sender.py
```diff
@@ -266,18 +298,22 @@ def test_upload_sender_post_called_with_right_parameters_test_results(
     ):
         headers = {"Authorization": f"token {random_token}"}
 
-        mocked_legacy_upload_endpoint.match = [
-            matchers.json_params_matcher(request_data),
-            matchers.header_matcher(headers),
-        ]
-
```
calvin-codecov, line 273: This seems like it was incorrectly using the mocked_legacy_upload_endpoint even though this is a different endpoint. I think we want the mocked_test_results_endpoint (newly added below) which would match with the argument at line 295 and with the url designated in this function

codecov-cli/codecov_cli/services/upload/upload_sender.py
```diff
@@ -92,11 +91,14 @@ def send_upload_data(
                     commit_sha,
                     report_code,
                     upload_coverage,
+                    file_not_found=file_not_found,
```
calvin-codecov, line 95: This seemed like a missed line from before. We weren't passing in the file_not_found value that is defined higher up at line 59/61 and self.get_url_and_possibly_update_data() takes it in with a default value of "False" so it would always just set `data['file_not_found'] to False

Please review this in particular

codecov-cli/codecov_cli/services/upload/upload_sender.py
```diff
@@ -92,11 +91,14 @@ def send_upload_data(
...
                 )
                 # Data that goes to storage
                 reports_payload = self._generate_payload(
                     upload_data, env_vars, report_type
                 )
+                reports_payload_size = len(reports_payload)
```
thomasrockhu-codecov, line 101: have you compared this to what comes out of `sys.getsizeof`?

thomasrockhu-codecov (approved):
Just a note to compare the file sizes

codecov-cli/codecov_cli/services/upload/upload_sender.py
calvin-codecov: I just tried it with example python and there was a slight difference.
`len`: 2432 bytes
`sys.getsizeof`: 2469 bytes

PR #738, opened 2026-07-09
fix: build x86_64 linux binary on glibc 2.31

thomasrockhu-codecov:
Build codecovcli_linux in the ubuntu:20.04 Docker container (glibc 2.31) instead of on the ubuntu-22.04 runner (glibc 2.35). PR #725 dropped the Dockerized x86_64 build, causing the binary to require glibc 2.35 and break on older distros such as AL2023 (glibc 2.34).

### DIFF
--- a/.github/workflows/build_assets.yml
+++ b/.github/workflows/build_assets.yml
@@ -29,13 +29,6 @@ jobs:
               lipo -archs dist/codecovcli_macos | grep 'x86_64 arm64'
             OUT_FILE_NAME: codecovcli_macos
 
-          - os: ubuntu-22.04
-            TARGET: ubuntu
-            CMD_BUILD: >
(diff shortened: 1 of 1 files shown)


calvin-codecov (approved):

.github/workflows/build_assets.yml
```diff
@@ -116,14 +113,14 @@ jobs:
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: codecovcli_${{ matrix.distro_name }}_${{ matrix.arch }}
-          path: ./dist/codecovcli_${{ matrix.distro_name }}_${{ matrix.arch }}
+          path: ./dist/codecovcli_*
```
calvin-codecov, line 116: These could be mismatched now making it harder to discern if we were to debug?

